PR #48, opened 2016-05-21
Allow Browser Sync port to be specified from command line

aarondfrancis:
Defaults to the default Browser Sync port: 3000.


mattstauffer:
:+1: Thanks Aaron!

devcircus:
Was about to take a shot at my first PR! Thanks for this!

Issue #51, opened 2016-06-07
Enable Elixir versioning


mattstauffer:
https://laravel.com/docs/elixir#versioning-and-cache-busting

Will require bringing in the `elixir()` helper and making sure it's configured correctly for source and target directories.

joshwhatk:
I would love to see this happen. I tried setting it up, but Jigsaw kept choking on the elixir() call.

For now, I've just added an extra step to my production build where I update file names. ¯\_(ツ)_/¯

joshwhatk:
I finally got it to work, I'm going to submit a pull request for it 😄

PR #54, opened 2016-06-16
Add Elixir versioning

joshwhatk:
Hey Adam,

As mentioned in #51 I was able to set up versioning for jigsaw so it works the same as it does in Laravel.

I included `src/Support/helpers.php` which contains `public_path()` and `elixir()`. `public_path()` makes sure that it uses the `rev-manifest.json` within the source folder. And I included it all in the files section of `composer.json` for autoloading.

I also noticed that it included a duplicate slash if an empty build path was used (which is necessary when using relative paths to fonts or images from css), so I included a check to handle that.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,10 @@
     "autoload": {
         "psr-4": {
             "TightenCo\\Jigsaw\\": "src"
-        }
+        },
+        "files": [
+            "src/Support/helpers.php"
+        ]
     },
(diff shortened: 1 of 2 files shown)


mariordev:
This will be really nice to have. Thanks @joshwhatk!

besologic:
@joshwhatk Thanks for your contribution!

Issue #56, opened 2016-06-25
[Feature] Allow for generating files with endings other than `.html`


mattstauffer:
Allow Jigsaw to generate files with non-`.html` endings so folks can generate `.js` files or `.rss` files.

Reference: #46 #50

damiani:
Support for this added in v1.0.0; you can now process `.js`, `.json`, `.xml`, `.rss`, `.txt`, and `.text` with Blade first: http://jigsaw.tighten.co/docs/other-file-types/

tnorthcutt:
Just for anyone else arriving here from the googles, the updated docs link is http://jigsaw.tighten.co/docs/content-other-file-types/

raniesantos:
Does this include blade.yml?

@tnorthcutt @damiani

damiani:
No, the only Blade hybrid files currently supported are those listed above and in the docs. Feel free to PR, though; it would require simple changes to the Blade extension arrays in `ViewRenderer` and `InputFile`, I believe.

Issue #46, opened 2016-05-09
RSS Feed


appkr:
I recently built a static site with jigsaw. In comparison with other static site builder, one missing feature was RSS feed generation. 

I did it manually using large data array in config.php and blade like the following..

```php
// config.php

return [
    'lessons' => [
        [
            'title' => 'Foo title',
            'url' => '/foo.html',
            'description' => 'Foo description...',
        ],
        [...],
    ],
];
```

```xml
<!--// feed.xml.blade.php -->

<?xml version="1.0" encoding="UTF-8"?>
<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
  <channel>
    <title>{{ $site_title }}</title>
    <description>{{ $site_description }}</description>
    <link>http://example.com</link>
    <atom:link href="http://example.com/feed.xml" rel="self" type="application/rss+xml"/>
    <pubDate>{{ (new DateTime)->format(DATE_ATOM) }}</pubDate>
    <lastBuildDate>{{ (new DateTime)->format(DATE_ATOM) }}</lastBuildDate>
    <generator>Blade templating engine</generator>

    @foreach($lessons as $lesson)
      <item>
        <title>{{ $lesson['title'] }}</title>
        <description>{{ $lesson['description'] }}</description>
        <pubDate>{{ (new DateTime)->format(DATE_ATOM) }}</pubDate>
        <link>http://example.com{{ $lesson['url'] }}</link>
        <guid isPermaLink="true">http://example.com{{ $lesson['url'] }}</guid>
      </item>
    @endforeach
  </channel>
</rss>
```

and then, in my build script

```sh
# build.sh

[ -f build_local/feed.xml.html ] && mv build_local/feed.xml.html build_local/feed.xml
```

No clean though, I was able to get a valid feed.xml.

While I'm at it, I'd like to ask one more. I hope I can access the array of the file to be compiled, so that I don't have to define the large data in config.php. Metadata like `title`, `description`, `published_at`, `...` may be read from front formatter of the markdown file.

paslandau:
I have excatly the same issue. My first attempt was to reegineer the `jigsaw` command and make the componentes (Jigsaw, the DI-Containert, ...) available in my Blade template but then I realized that they don't expose the meta data from my mark down posts via public methods.

At that point it occured to me that I might not be the only one with that problem and before it got even messier I decided to check github :)

[...]

PR #66, opened 2016-09-10
Allow for generating files with endings other than `.html`

TheNodi:
Referencing issue #56.

This allow any file in the format `abc.blade.xxx` to be compiled to `abc.xxx` using blade template.

It was necessary to overwrite view factory because the compile engine is selected based on file extension and it's hard-coded in `$this->extensions` property. It would be possible to extend extensions property by end user to map their custom extension to blade engine, but I think compiling every .blade.xxx file is much easier and cleaner.

Example (using #46 ).

File `source/feed.blade.xml`:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
    <channel>
        <title>{{ $site_title }}</title>
        <description>{{ $site_description }}</description>
        <link>http://example.com</link>
        <atom:link href="http://example.com/feed.xml" rel="self" type="application/rss+xml"/>
        <pubDate>{{ (new DateTime)->format(DATE_ATOM) }}</pubDate>
        <lastBuildDate>{{ (new DateTime)->format(DATE_ATOM) }}</lastBuildDate>
        <generator>Blade templating engine</generator>
    </channel>
</rss>
```

Will be compiled to `build_local/feed.xml`:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
    <channel>
        <title>Hello World</title>
        <description>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque mattis vestibulum est ut facilisis. In porta quis diam in efficitur. Donec ligula tellus, sagittis id nunc eu, iaculis ullamcorper neque. </description>
        <link>http://example.com</link>
        <atom:link href="http://example.com/feed.xml" rel="self" type="application/rss+xml"/>
        <pubDate>2016-09-10T22:17:38+02:00</pubDate>
        <lastBuildDate>2016-09-10T22:17:38+02:00</lastBuildDate>
        <generator>Blade templating engine</generator>
    </channel>
</rss>
```

### DIFF
--- a/jigsaw
+++ b/jigsaw
@@ -17,11 +17,12 @@ use TightenCo\Jigsaw\Handlers\DefaultHandler;
 use TightenCo\Jigsaw\Handlers\MarkdownHandler;
 use TightenCo\Jigsaw\Jigsaw;
 use TightenCo\Jigsaw\TemporaryFilesystem;
+use TightenCo\Jigsaw\ViewFactory;
 
-if (file_exists(__DIR__.'/vendor/autoload.php')) {
-    require __DIR__.'/vendor/autoload.php';
+if (file_exists(__DIR__ . '/vendor/autoload.php')) {
+    require __DIR__ . '/vendor/autoload.php';
 } else {
-    require __DIR__.'/../../autoload.php';
+    require __DIR__ . '/../../autoload.php';
 }
 
 // Config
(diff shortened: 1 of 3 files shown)


erickpatrick:
This one as well can be clased as we have this functionality on master

damiani:
Thanks for contributing this! Support for this was added in v1.0.0, but it required a slightly different approach because of big changes from v0.6.4 in how files are parsed. You can now process `.js`, `.json`, `.xml`, `.rss`, `.txt`, and `.text` with Blade first: http://jigsaw.tighten.co/docs/other-file-types/

PR #67, opened 2016-09-11
Add ability to register custom blade tags

TheNodi:
Referencing issue #20.

To register a custom blade directive you just create a file in the `/blade` directory, if you create `/blade/datetime.php` it will register the `@datetime` blade directive.

There's an important difference from Laravel's way of registering directives. The default way doesn't needs you to return the php code to put in your compiled view.
## Simple Directive (Default)

Example:

``` php
// File: /blade/datetime.php

/** @var $input mixed */
/** @var $parameters array */

echo date('Y-m-d H:i:s', $input);
```

This php file will be executed at runtime, receiving the evaluated object as `$input`. I found myself not needing to specify anything more complex then this in the majority of cases.

`$parameters` will be an array of current views parameters, which allows you to act based on configurations variables like `if ($parameters['production']) ...`.

The directive file has its own variable scope, which means it's possible to create and manipulate any variable without interfering with view's variables.
## Advanced Directive (Laravel Style)

If you need to work with the input expression (like Laravel's [extending feature](https://laravel.com/docs/5.3/blade#extending-blade)), you easily can just by naming the file `.raw.php`.

To replicate the datetime example in the previous step, you can do something like this:

``` php
// File: datetime.raw.php

/** @var $expression string */

return '<?php echo date("Y-m-d H:i:s", ' . $expression . '); ?>';
```

I've registered the service container using `Container::setInstance($container);` so that people can get the instance from their blade files. In the future they might be able to bind their own object in the service container.

### DIFF
--- a/jigsaw
+++ b/jigsaw
@@ -28,14 +28,36 @@ if (file_exists(__DIR__.'/vendor/autoload.php')) {
 $cachePath = getcwd() . '/_tmp';
 $buildPath = getcwd() . '/build';
 $sourcePath = getcwd() . '/source';
+$bladePath = getcwd() . '/blade';
 
 $container = new Container;
 
-$container->bind(Factory::class, function ($c) use ($cachePath, $sourcePath) {
+$container->bind(Factory::class, function ($c) use ($cachePath, $sourcePath, $bladePath) {
     $resolver = new EngineResolver;
 
-    $resolver->register('blade', function () use ($cachePath) {
+    $resolver->register('blade', function () use ($cachePath, $bladePath) {
         $compiler = new BladeCompiler(new Filesystem, $cachePath);
+
+        if (is_dir($bladePath)) {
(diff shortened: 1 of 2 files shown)


damiani:
This is a great idea, but I think it would be clearer if (1) we only go with one approach, more like your "advanced" example, and (2) we make the API similar to the way [helper methods](http://jigsaw.tighten.co/docs/helper-methods/) are handled in v1.0.0. 

So instead of using separate files for each directive stored in a `blade` directory, maybe we could use a `blade.php` config file, similar to `config.php`, that returns an array of closures for each directive, keyed by the directive name, like this:

```php
<!--  blade.php -->

return [
    'datetime' => function($expression, $page) {
        return '<?php echo date("Y-m-d H:i:s", ' . $expression . '); ?>';
    },
];
```

...with `$page` doing the same thing that `$parameters` does in your "simple directive" example.

TheNodi:
See #95

PR #68, opened 2016-09-11
Customizing Parsedown

TheNodi:
Referencing issue #64.

Placing a `bootstrap.php` file on the root of your jigsaw will load it before registering all file handlers.
The markdown and yaml parsers are resolved using the service container so you can re-bind them and extend parsedown.

## Example

Files to make the [inline colored text](https://github.com/erusev/parsedown/wiki/Tutorial:-Create-Extensions) in the parsedown wiki.

```php
<?php
// File: /bootstrap.php

use App\UberParsedown;
use Mni\FrontYAML\Markdown\MarkdownParser;

/** @var $container \Illuminate\Container\Container */
/** @var $jigsaw \TightenCo\Jigsaw\Jigsaw */

$container->bind(MarkdownParser::class, UberParsedown::class);
```

```php
<?php
// File: /app/UberParsedown.php

namespace App;

use Mni\FrontYAML\Markdown\MarkdownParser;
use Parsedown;

class UberParsedown extends Parsedown implements MarkdownParser
{
    function __construct()
    {
        $this->InlineTypes['{'] [] = 'ColoredText';

        $this->inlineMarkerList .= '{';
    }

    protected function inlineColoredText($Excerpt)
    {
        if (preg_match('/^{c:([#\w]\w+)}(.*?){\/c}/', $Excerpt['text'], $matches)) {
            return array(
                'extent' => strlen($matches[0]),
                'element' => array(
                    'name' => 'span',
                    'text' => $matches[2],
                    'attributes' => array(
                        'style' => 'color: ' . $matches[1],
                    ),
                ),
            );
        }

        return null;
    }
}
```

Now your markdown file:
```markdown
---
extends: _layouts.test_layout
section: content
---

# My awesome heading!

{c:#FFF}My awesome content!{/c}
```

Will be rendered as:
```html
<h1>My awesome heading!</h1>
<p><span style="color: #FFF">My awesome content!</span></p>
```

This has the side-effect to allow any further customization, like registering custom JigSaw handlers.

### DIFF
--- a/jigsaw
+++ b/jigsaw
@@ -8,6 +8,9 @@ use Illuminate\View\Engines\CompilerEngine;
 use Illuminate\View\Engines\EngineResolver;
 use Illuminate\View\Factory;
 use Illuminate\View\FileViewFinder;
+use Mni\FrontYAML\Markdown\MarkdownParser;
+use Mni\FrontYAML\Parser;
+use Mni\FrontYAML\YAML\YAMLParser;
 use TightenCo\Jigsaw\Console\BuildCommand;
 use TightenCo\Jigsaw\Console\InitCommand;
 use TightenCo\Jigsaw\Console\ServeCommand;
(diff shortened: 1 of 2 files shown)


jigsaw
```diff
@@ -47,13 +51,23 @@ $container->bind(BladeHandler::class, function ($c) {
     return new BladeHandler($c[Factory::class]);
 });
 
+$container->bind(MarkdownParser::class, \Mni\FrontYAML\Bridge\Parsedown\ParsedownParser::class);
+$container->bind(YAMLParser::class, \Mni\FrontYAML\Bridge\Symfony\SymfonyYAMLParser::class);
```
damiani: For consistency, let's import these classes in the `use` block and just refer to the final class names `ParsedownParser::class` and `SymfonyYAMLParser::class` here.

jigsaw
```diff
@@ -47,13 +51,23 @@ $container->bind(BladeHandler::class, function ($c) {
...
+$container->bind(Parser::class, function ($c) {
+    return new Parser($c[YAMLParser::class], $c[MarkdownParser::class], '---', '---');
+});
```
damiani: The two separators `'---'` aren't necessary, FrontYAML defines those by default in its constructor.

damiani:
I really like this solution! 

- The changes made to the `jigsaw` file will need to be updated soon because that file is changing substantially in an impending update, but we'll port these changes over once that becomes necessary.

- I also think it would be a good idea to add psr-4 loading of `App` to `composer.json` so people who implement this don't need to muck about in there.

TheNodi:
I've fixed the consistency errors.

There's no `composer.json` file scaffold with init command to add the psr-4 loading feature. I've create the App namespace in my own application and used it as an example, but maybe it could enable a better organization of extensions (like blade file location in PR #67 ). Maybe the init command can scaffold a `composer.json` file if not already available (local installation).

damiani:
Very true; for now, we can just cover it in the docs, and will consider adding a scaffolded `composer.json` in the future.

Thanks for your contribution!

PR #75, opened 2017-01-21
Ability to get Collection object from CollectionItem object

fideloper:
When re-using templates for multiple collections, and you don't know the variable name of the collection (which I suppose is itself might be a trick to document?), this lets you get access to the collection from an `$item`.

I used this so I could `$item->getCollection()->getPermalink()` to create a link back to the "root" of the collection (e.g. blog listing page). For next/back links, `$item->getNext()`, etc, were sufficient, but this may let us do more tricks with collections for more custom uses.

### DIFF
--- a/src/Collection/CollectionItem.php
+++ b/src/Collection/CollectionItem.php
@@ -16,6 +16,11 @@ class CollectionItem extends IterableObject
 
         return $item;
     }
+    
+    public function getCollection()
+    {
+        return $this->collection;
+    }
 
     public function getHelper($name)
     {


damiani:
My concern is that adding something like `getCollection()` would expose a bunch of methods on the `Collection` object that aren't meant to be user-facing, and that can have unpredictable results. `getPermalink()`, for instance, isn't really returning the "root" path per se, it's returning the `path` setting defined in `collections.php`, which could just as well be a string to be parsed (like `posts/{date|Y-m-d}/{+filename}`, or an array of paths, or even a function. The root path for a collection is actually dependent on the filename of whatever template you choose to use as an index page, something like `posts.blade.php`; but the collection doesn't actually have any knowledge of that as a root.

You _can_ already get access to the _name_ of the collection from an item, using `$item->collection`; in addition, you could get access to the whole collection from within an item using the (rather ugly) `${$item->collection}` without having to know the collection's name up front. So `${$item->collection}->first()` would give you the same results as `$item->getFirst()`.

But for something like a root path, it would be safer to set that up as a variable in `collections.php`:

```
'variables' => [
    'root' => '/all-my-blog-posts',
],
```

...and consume that in your template as `$item->root`.

Can you think of other use cases where you would need access to the collection from within an item, or of other helper methods like `getFirst()`, `getNext()`, etc. that would be useful?

fideloper:
Thanks for clarifying, makes sense. Adding the variable to the collection
will work great.

erickpatrick:
We also have this on master, as per #79, by using `$page`

damiani:
In v1.0.0, you can call `$page->getCollectionName()` to return the name of your collection, and build a root link from that.

Or, you can use the variable solution mentioned above, though the syntax has changed; it would go in `config.php` under the key for your collection, and you no longer need the `variables` key:

```php
<?php

return [
    'collections' => [
        'posts' => [
            'root' => '/blog',
        ],  
        'people' => [
            'root' => '/who-we-are',
        ],      
    ],
];

PR #76, opened 2017-01-24
Add specifying custom build destination

vmitchell85:
Added an option to specify a different folder destination name

### DIFF
--- a/src/Console/BuildCommand.php
+++ b/src/Console/BuildCommand.php
@@ -24,7 +24,8 @@ class BuildCommand extends Command
         $this->setName('build')
             ->setDescription('Build your site.')
             ->addArgument('env', InputArgument::OPTIONAL, "What environment should we use to build?", 'local')
-            ->addOption('pretty', null, InputOption::VALUE_REQUIRED, "Should the site use pretty URLs?", 'true');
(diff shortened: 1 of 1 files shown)


erickpatrick:
And, once again, we have this one on master. This can be closed.

vmitchell85:
@erickpatrick Thanks for the info... I found it in an old PR but not in the docs, which I searched for before creating this PR.

I think I'd still prefer the command line option though, instead of specifying in config.php.

However I hear Jigsaw is undergoing a lot or changes so I don't think this will matter much.

damiani:
@vmitchell85 We had several people express a need for customizing the source/destination locations, such that making it a permanent `config` option—which could also be customized by environment—seemed like a user-friendly and worthwhile feature. I left this open, though so that once the `master` branch is tagged as a release, we could discuss updating your PR to add it as a command line option as well. I can see valid use cases for both approaches.

damiani:
I think we'll stick with the `config`-based setting for now, haven't encountered a compelling reason to add it as a command line option.

vmitchell85:
Thanks for taking the time to consider it.

I've changed my process since the rewrite to use the config option... so it's no longer an issue for me.

PR #83, opened 2017-03-14
Replace deprecated quickRandom() with random()

blueclock:
quickRandom is deprecated in 5.3


damiani:
Good to know, thanks!

PR #85, opened 2017-03-23
Allow gulp command to check which jigsaw binary to use

erickpatrick:
When running gulp, instead of making the user change the gulp file, manually,
add a check for which jigsaw binary to use (global or local under vendor
folder) and in the case none exists prints a message to stdout via
console.log and exits the process.

### DIFF
--- a/site/gulpfile.js
+++ b/site/gulpfile.js
@@ -1,6 +1,7 @@
 var gulp = require('gulp');
 var elixir = require('laravel-elixir');
 var argv = require('yargs').argv;
+var fs = require('fs')
 
 elixir.config.assetsPath = 'source/_assets';
 elixir.config.publicPath = 'source';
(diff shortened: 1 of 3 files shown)


damiani:
Fantastic! Could you add a little bit to the console error message, to mention using Composer? Something like `Could not find Jigsaw; please install it via Composer, either locally or globally.`

👍

erickpatrick:
Sure, I'll work on that =)

erickpatrick:
Hi @damiani, please, check the changes I made. After some testing I decide to change how I approached this, but now it seems to be working better than the previous one. [See the commit message](https://github.com/tightenco/jigsaw/pull/85/commits/7b7ec21693aa7ab826ebf5982cdb8af9c457cc30).

Also, I added the message you asked.

damiani:
Looks great, though I believe we don't need the added `fs` dependency in `package.json`, since `fs` is part of the core.

erickpatrick:
Hi @damiani, removed the dependency. If you think it can be merged, let me know if you rather have it squashed into a single commit or if you're going to do that by GitHub interface.

damiani:
Excellent, thanks for this!

erickpatrick:
@damiani my pleasure! Expect more to come, I'm really enjoying working with jigsaw

PR #87, opened 2017-03-28
Ignore folders and files generated for elixir versioning/cache busting

erickpatrick:
When using Elixir's ability to version/cache bust assets, it generates
folders and files inside `source` making `jigsaw` check its content
to generate HTML files.

By excluding those folders from jigsaw command, we avoid unnecessary
work. By also adding them to `.gitignore` we don't pollute the repo
versioning with generated files that will be sent to `build_<env>`
in the end.

For further infor on Elixir's versioning/cache busting functionality
we can refer to [Elixir's documentation](https://laravel.com/docs/5.3/elixir#versioning-and-cache-busting)

### DIFF
--- a/site/.gitignore
+++ b/site/.gitignore
@@ -1,3 +1,6 @@
 /build_local/
 /node_modules/
 /vendor/
+/source/build
+/source/css
+/source/js
--- a/site/gulpfile.js
+++ b/site/gulpfile.js
@@ -11,7 +11,7 @@ elixir(function(mix) {
     var port = argv.p || argv.port || 3000;
 
     mix.sass('main.scss')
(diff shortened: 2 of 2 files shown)


damiani:
Since we don't `version` by default in `gulpfile.js`, I feel like excluding `/source/build` by default isn't necessary and could be confusing to someone who makes a `/build` directory that they expect to get processed. Someone who adds `version` to the Elixir process could just exclude it then.

We also wouldn't want to `gitignore` `/source/css` or `/source/js`, for cases where people are writing js or css files without using a preprocessor.

erickpatrick:
Great, so I'll close this PR so it does not pile up the list, thanks =)

Issue #86, opened 2017-03-27
Browsersync/Gulp watch choking on many files


cjf:
jigsaw: 0.6.4
PHP 7.1.2

Running 'gulp watch' has become tedious now that my static site is approaching 1k "posts".  Any change seems to be triggering several browsersync "browser reloads" for each file which results in instability for the utility. 

I'm fairly new to both gulp and browsersync so maybe I'm misunderstanding what is happening, but it seems like browsersync is triggering as jigsaw builds the site instead of doing it just once after jigsaw is finished.

In any event, I've attached a simple demo project that has 1000 "posts".  It doesn't choke the whole thing down due to the simplicity of the content, but it does demonstrate the reloading problem and you can hopefully imagine that a more complex site may be more crippled by this process.

Is there something that can be done to force the "watch" mechanism to wait until jigsaw is done building before "checking" it?

[issue.zip](https://github.com/tightenco/jigsaw/files/873446/issue.zip)

damiani:
I haven't looked into Elixir closely to see why this would be happening, but an easy fix might be to use Browsersync's `reloadDebounce` option—which waits for a specified period of inactivity before reloading the browser. In order to use this, though, you need to first update `laravel-elixir`, because the version of Elixir currently used by Jigsaw in v0.6.4 doesn't fully support `reloadDebounce.`

So, in your `package.json`, remove the `laravel-elixir` line, and replace it with:

```
    "laravel-elixir": "^6.0.0-15",
    "laravel-elixir-browsersync-official": "^1.0.0"
```

then run `npm install` from your project root.

Then, in `gulpfile.js`, add this option to the `.browserSync({...})` step:

```
reloadDebounce: 2000
```

That will pause for 2 seconds before trying to reload, giving a long build process some time to settle down, so Browsersync doesn't freak out with all the file changes. You can adjust the debounce value as needed.

Try it out, and let me know if it fixes the issue.

cjf:
Just tried it out.  While it does keep BS from freaking out, it also seems to break it.  That is, the browser connection gets dropped, so no reloading happens at all.  I've tried various values for reloadDebounce (from 1 to 10000): they all have the same effect.

I did a bit of googling around to see if anyone else has the issue of the connection dropping, but I can't find anything of use.

For what it's worth, if I reload localhost:3001 to see the connections after a BS event, the connection that existed before, disappears.

cjf:
Also, the BS code snippet is injected in the page source.  I can see it.

[...]

PR #89, opened 2017-04-05
Switch from Laravel Elixir to Laravel Mix

erickpatrick:
Even though Laravel Elixir was recently updated to its latest version
on this project it is a EOL module. The new tool, specially for those
familiar to Laravel, is Laravel Mix.

With this Pull Request it's providen a full working version of Jigsaw
using Laravel Mix instead of Laravel Elixir.

It's important to mention that Mix surpasses Elixir, providing much
more functionality. The only exception to this is: `.exec`. Thus we
use a Webpack plugin that is able to call external commands so we
are able to call Jigsaw. The call to Jigsaw still complies with
issue #85, where it is called from local or global composer,
whichever is available.

Now, `npm run webpack` is the equivalent of running `gulp` and
`npm run dev` is the one for `gulp watch`. Both of them run for 
local builds (`--env=local`). To run the production version we 
can use `npm run production`. There are other commands 
as well, coming from Mix, all of them on `package.json`.


erickpatrick:
Please, if possible, test on big projects using Jigsaw as the ones I work on are small (+- 50 pages, +- 300 posts/collection items)

damiani:
Just the PR I was hoping for...I'll put this through its paces and make sure it works with larger sites.

erickpatrick:
Added new commit to remove `gulpfile.js` as it is not needed anymore

arsenx:
Thanks so much for this @erickpatrick 😄 

I was wondering why I am not able to run `mix.copyDirectory('source/_assets', 'build_local/assets');`
Any ideas on this? Thanks!

damiani:
@erickpatrick This seems to break browserSync support for me, in that running the `dev` script doesn't seem to be successfully watching for changes and rebuilding. It's also choking on `.less` files, seems to be missing a loader.

It would also be best if the scripts in `package.json` were in sync with the current setup on Laravel 5.4 (https://github.com/laravel/laravel/blob/master/package.json), i.e. change `webpack` -> `dev`, `dev` -> `watch`, etc.

erickpatrick:
@damiani I'll have a look on this. Thanks for the feedback.

arsenx:
Thanks so much for this PR @erickpatrick! I've been testing it out and works great!
I'd like to contribute by adding the asset() helper feature from laravel into Jigsaw, any ideas on how we could achieve that?

erickpatrick:
Hi guys, my agenda has been a full for some time, but soon(ish) I'll get back to this and try to make it work as expected (given the feedback)

JonasDoebertin:
Would be even nicer, if Laravel Mix asset versioning would be supported 👍

site/webpack.mix.js
```diff
@@ -0,0 +1,51 @@
+let mix = require('laravel-mix');
+let argv = require('yargs').argv;
+let bin = require('./tasks/bin');
+let WebpackShellPlugin = require('webpack-shell-plugin');
+
+let env = argv.e || argv.env || 'local';
+var port = argv.p || argv.port || 3000;
```
PascaleBeier: Shouldn't this rather say 'let env'?

afontcu: Actually, my two cents, I'd use `const`, since these variables (mix, argv, binc, WebpackShellPlugin...) are not meant to change, and since there's a debate about the proper use of let and const, I'd say this is the most rational approach:

https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75

> `const` is a signal that the identifier won’t be reassigned.
`let` is a signal that the variable may be reassigned, such as a counter in a loop, or a value swap in an algorithm.

If this has already been discussed and I'm just opening the Pandora Box again... ignore me :)

Thanks!

PascaleBeier: @afontcu You are right, but I mostly tried to stay consistent with the Mix Codestyle which is using let in these cases as well. 

Really seems to be the Box of Pandora 😄

JonasDoebertin:
I toyed around a bit getting Laravel Mix running with Jigsaw and came to a quite decent solution.

```js
let mix = require('laravel-mix');
let argv = require('yargs').argv;
let jigsaw = require('./tasks/bin');
let command = require('node-cmd');

let WatchPlugin = require('webpack-watch');
let BeforeBuildPlugin = require('before-build-webpack');

const env = argv.e || argv.env || 'local';
const port = argv.p || argv.port || 3000;
const buildPath = 'build_' + env + '/';

let plugins = [
    new WatchPlugin({
        paths: ['source/**/*.md', 'source/**/*.php'],
        options: {}
    }),

    new BeforeBuildPlugin((compiler, proceed) => {
        command.get(jigsaw.path() + ' build ' + env, (error, stdout, stderr) => {
            console.log(error ? stderr : stdout);
            proceed()
        });
    })
];

mix.webpackConfig({ plugins });

mix.setPublicPath(buildPath + 'assets/');

mix.js('source/_assets/js/main.js', '/js/')
    .sass('source/_assets/sass/main.scss', '/css/');

mix.browserSync({
    port: port,
    server: { baseDir: buildPath },
    proxy: null,
    files: [ buildPath + '**/*' ]
});

```

This works with running single compilations `npm run dev` as well as starting watchers `npm run watch`. The assets are being compiled/placed right into the build directory. This can be done because a Jigsaw build will be triggered each time before asset compilation starts. Also it will watch for any changed `.md` or `.php` files in the `source` directory and will trigger a build/compilation once files are beeing changed.

You'll need the `node-cmd`, `webpack-watch` and `before-build-webpack` packages though.

AyubM:
It should be acknowledged, webpack (Laravel Mix) and gulp (Elixir) are not entirely the same things.

Given the nature of static site generators, would should keep gulp as the default. Gulp is itself is a task runner / generator, and can be used as an extension to jigsaw.

For example, there are pull requests for jigsaw like "remote api collections" ... this could be solved by the developer by writing a gulp task to fetch his remote articles and load them up into his collection folder. There's no need to extend jigsaw for this and even less need to merge this code in for everyone.

Gulp lets developers bend jigsaw to their needs. There are all kinds of ways we can use Gulp to load data and enhance jigsaw with our scripts that we couldn't do with webpack. If we changed jigsaw to webpack, then anyone who had custom requirements would also needlessly have to install gulp side by side.

We don't have to follow Laravel on their decision to move to webpack. We can do what's best for the user of jigsaw and stick with one system that makes sense for the majority of use cases jigsaw fills -- no one's building REST SPA's on top of jigsaw to warrant webpack. A gulp script is fine for almost all jigsaw compilation needs, plus it lets you extend functionality should you need it.

damiani:
These are good points. I'm interested in using Webpack not for Webpack's power _per se_, but because I'm more concerned about remaining dependent on Elixir once it's EOL'd.

At least theoretically, much of the task-running behavior of Gulp (including the remote API example) could be accomplished using NPM scripts. Other than the API example, can you describe some other custom tasks that would be made more difficult or impossible with an NPM script?

damiani:
@cmosguy Have you only encountered failures on Windows?

damiani:
I've pushed some updates that build on @JonasDoebertin's approach, with some important changes:

- `jigsaw build` is triggered _after_ the webpack build is complete, not before, otherwise we can't use the `{{ mix() }}` helper in a Blade file to version assets. This means we still have send the results of asset compilation to an intermediate `assets` directory in `source`, and then copy that to `build` as before.
- BrowserSync is working, not using Mix's `mix.browserSync` command, but instead by initializing it manually so we can get a handle to the instance. Using this handle, we can trigger `reload()` after webpack is done compiling assets and Jigsaw is done compiling files, which seems to prevent BroswerSync from running prematurely, getting caught in loops, etc.
- I removed the `files: [ buildPath + '**/*' ]` option from the BrowserSync config, which eliminates the flood of `[Browsersync] Reloading Browsers...` messages when rebuilding; previously, one message would appear for each file being watched. This may be just a cosmetic change, or it may actually be more efficient to just watch the whole directory—I haven't dug into it to figure out which is the case. I'm crossing my fingers that this might solve https://github.com/tightenco/jigsaw/issues/86

Note that this requires more recent versions of Mix and BrowserSync, so `package.json` has been updated.

@JonasDoebertin and @erickpatrick, can you test this out and report back? @cmosguy, can you give this a spin on Windows and see if it works?

darol100:
👍  For Laravel mix

cmosguy:
@damiani I just attempted to test his PR with your edits, I marked a comment with the `report()` and `app()` global methods.  Besides that, the `mix()` method is almost working.  So I have tested on windows.

damiani:
@cmosguy If you haven't changed anything from the default `webpack.mix.js` settings [here](https://github.com/tightenco/jigsaw/pull/89/files#diff-07309cf1a12e3022342fb4729cd3d1d6), then you should be able to link to a JS file in `master.blade.php` by either:

```
<script src="assets/js/main.js"></script>
```

or

```
<script src="{{ mix('js/main.js') }}"></script>
```

Running `npm run dev`, and using the default settings in `webpack.mix.js`, will compile the JS/SASS assets (from the `source/_assets` directory) and output them to a `source/assets` directory. Then, that `assets` directory in `source` gets copied to the `build_local` (or `build_production`) directory. So after webpack is run and the site is built, you should end up with this:

```
├── build_local
│   ├── assets
│   │   ├── css
│   │   │   └── main.css
│   │   ├── js
│   │   │   └── main.js
│   │   └── mix-manifest.json
│   └── index.html
└── source
    ├── _assets
    │   ├── js
    │   │   └── main.js
    │   └── sass
    │       └── main.scss
    ├── _layouts
    │   └── master.blade.php
    ├── assets
    │   ├── css
    │   │   └── main.css
    │   ├── js
    │   │   └── main.js
    │   └── mix-manifest.json
    └── index.blade.php
```

Log1x:
Surely this Mix configuration can be simplified...

How about something like this:

```js
const mix     = require('laravel-mix');
const jigsaw  = require('./tasks/bin');
const command = require('node-cmd');

/*
 |--------------------------------------------------------------------------
 | Mix Asset Management
 |--------------------------------------------------------------------------
 |
 | Mix provides a clean, fluent API for defining some Webpack build steps
 | for your Laravel application. By default, we are compiling the Sass
 | file for the application as well as bundling up all the JS files.
 |
 */

const app       = 'source';
const build     = 'build_local'
const assets    = `${app}/assets`;
const dist      = `${app}/dist`;

mix.setPublicPath(dist);
mix.setResourceRoot('../');

// Browser Sync
mix.browserSync({
  proxy: 'https://jigsaw.dev',
  browser: ['firefox.exe'],

  files: [
    `${build}/**/*.php`,
    `${build}/**/*.md`,
    `${dist}/styles/**.css`,
    `${dist}/scripts/**.js`
  ]
});

// SASS
mix.sass(`${assets}/styles/main.scss`, `${dist}/styles/main.css`);

// Javascript
mix.js(`${assets}/scripts/main.js`, `${dist}/scripts/main.js`);

// Build Jigsaw
mix.then(function () {
  command.get(`${jigsaw.path()} build`, (error, stdout, stderr) => {
    console.log(error ? stderr : stdout);
  });
});
```

Log1x:
I think I made the best one yet..

```js
const mix = require('laravel-mix');
const bin = require('./tasks/bin');
const command = require('node-cmd');

/*
 |--------------------------------------------------------------------------
 | Mix Asset Management
 |--------------------------------------------------------------------------
 |
 | Mix provides a clean, fluent API for defining some Webpack build steps
 | for your Laravel application. By default, we are compiling the Sass
 | file for the application as well as bundling up all the JS files.
 |
 */

const app = 'source';
const build = 'build_local'
const assets = `${app}/assets`;
const dist = `${app}/dist`;

// Jigsaw
const jigsaw = () => {
  command.get(`${bin.path()} build`, (error, stdout, stderr) => {
    console.log(error ? stderr : stdout);
  })
};

// Options
mix.setPublicPath(dist);
mix.setResourceRoot('../');

// Browser Sync
mix.browserSync({
  host: 'localhost',
  port: 3000,
  proxy: 'https://jigsaw.dev',
  files: [
    `${build}/**/*.html`,
    `${dist}/styles/**.css`,
    `${dist}/scripts/**.js`,
    {
      match: [
        `${app}/**/*.php`,
        `${app}/**/*.md`,
      ],
      fn: () => jigsaw()
    }
  ],
});

// Assets
mix.sass(`${assets}/styles/main.scss`, `${dist}/styles/main.css`)
  .js(`${assets}/scripts/main.js`, `${dist}/scripts/main.js`)
  .then(() => jigsaw());

// Source maps when not in production.
if (!mix.inProduction()) {
  mix.sourceMaps();
}

`package.json`
```js
{
  "private": true,
  "scripts": {
    "build": "cross-env NODE_ENV=development webpack --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
    "build:production": "cross-env NODE_ENV=production webpack --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
    "start": "cross-env NODE_ENV=development webpack --progress --hide-modules --watch --watch-poll --config=node_modules/laravel-mix/setup/webpack.config.js",
    "start:hot": "cross-env NODE_ENV=development webpack-dev-server --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js"
  },
  "devDependencies": {
    "browser-sync": "^2.18.13",
    "browser-sync-webpack-plugin": "^1.2.0",
    "cross-env": "^5.1.1",
    "hasbin": "^1.2.3",
    "laravel-mix": "^1.6.1",
    "node-cmd": "^3.0.0"
  }
}
```

// Hash and version files in production.
if (mix.inProduction()) {
  mix.version();
}
```

damiani:
@Log1x Thanks, I'll test these out!

Log1x:
Updated my previous post with a few fixes/tweaks. It has been working quite well thus far.

damiani:
@Log1x I like the organization of this and reducing some of the external dependencies, but it brings back the earlier undesirable behavior (from https://github.com/tightenco/jigsaw/issues/86, and which you mention above) where a flood of `[Browsersync] Reloading Browsers...` is triggered, and where sometimes BrowserSync tries to reload before Webpack has finished asset processing. That's why I was manually calling `browserSyncInstance.reload()`. See the (comment above)[https://github.com/tightenco/jigsaw/pull/89#issuecomment-326806315]. It also leads to a much slower and "choppier" reload. 

Also, can you elaborate on what `mix.setResourceRoot('../');` is doing? I'm not familiar with that option, and it doesn't seem to have any effect in my testing.

PR #90, opened 2017-04-13
Pagination per page default option on config file

maicoqb:
### Added default_perPage option on config.php

This option allows the user to control the pagination globally on site.
Avoiding the default 10 items per page.


maicoqb:
Sorry guys, i did the PR in wrong way.
I will make another in a separate branch.

PR #92, opened 2017-04-17
Pagination per page default option on config file

maicoqb:
### Added default_perPage option on config.php

This option allows the user to control the pagination globally on site.
Avoiding the default 10 items per page.

### DIFF
--- a/src/Handlers/PaginatedPageHandler.php
+++ b/src/Handlers/PaginatedPageHandler.php
@@ -36,7 +36,7 @@ class PaginatedPageHandler
 
         return $this->paginator->paginate($file,
             $pageData->get($pageData->page->pagination->collection),
-            $pageData->page->pagination->perPage ?: ($pageData->page->default_perPage ?: 1)
+            $pageData->page->pagination->perPage ?: ($pageData->page->perPage ?: 10)
         )->map(function ($page) use ($file, $pageData) {
             $pageData->setPagePath($page->current);
             $pageData->put('pagination', $page);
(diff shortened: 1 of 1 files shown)


damiani:
Looks good, but let's change the variable name in `config.php` to just `perPage` so that it follows the same logic as cascading variables in v1. I would also keep the fallback as `10` rather than `1`.

maicoqb:
Seems good. 
I just use that name because I don't have another idea.
And about that, I think that `per_page` is a better name. As the snake_case is the patter on the others variables. But this will impact on the understanding, since the `perPage` is the current name on templates.

The fallback of 10 itens must be documented. I just notice that when I searched on the code.

maicoqb:
I did created a PR on jigsaw-siete adding on docs the default value of `perPage` when it's not defined.

https://github.com/tightenco/jigsaw-site/pull/13

src/Handlers/PaginatedPageHandler.php
```diff
@@ -36,7 +36,7 @@ public function handle($file, PageData $pageData)
 
         return $this->paginator->paginate($file,
             $pageData->get($pageData->page->pagination->collection),
-            $pageData->page->pagination->perPage ?: 10
+            $pageData->page->pagination->perPage ?: ($pageData->page->default_perPage ?: 1)
```
damiani: - Please change `default_perPage` to `perPage`
- Update the fallback default from `1` to `10`

damiani:
We'll update the docs to include the default of `10`, and we'll document this new default `perPage` setting in `config.php` as well, once this hits the next release. If you can make the two changes mentioned above, I can merge this in.

maicoqb:
@damiani , sorry, I did make these changes, but I forgotten to push to my repository.
Only on tuesday I will do that.

PR #93, opened 2017-04-17
Adding permalink option on page data

maicoqb:
That feature allows user to set a `permalink` key in the page's YAML front matter.

> 404.blade.md
``` markdown
---
extends: _layouts.master
title: "404 Error"
permalink: 404.html
section: body
---

### Oops! Maybe you did lost your path.
```

This configuration will change the default output from `404/index.html` to permalink's destination: `404.html`.

### DIFF
--- a/src/SiteBuilder.php
+++ b/src/SiteBuilder.php
@@ -98,11 +98,23 @@ class SiteBuilder
 
     private function getOutputDirectory($file)
     {
+        if( $permalink = $this->getFilePermalink($file) ) {
+            return urldecode(dirname($permalink));
+        }
+
         return urldecode($this->outputPathResolver->directory($file->path(), $file->name(), $file->extension(), $file->page()));
     }
 
     private function getOutputPath($file)
     {
+        if( $permalink = $this->getFilePermalink($file) ) {
+            return urldecode($permalink);
+        }
+
         return urldecode($this->outputPathResolver->path($file->path(), $file->name(), $file->extension(), $file->page()));
+    }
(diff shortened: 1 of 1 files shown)


damiani:
Can you elaborate a bit on why you need the `.html` extension in this case? 

It might make more sense to solve this by allowing `*.blade.html` files to be compiled to `*.html` pages, just like test-type files like `*.blade.js` and `*.blade.xml` are.

maicoqb:
Sometimes we need to use only one file with permalink, like int GitHub Pages' 404 error case.

GH Pages uses a 404.html file on site's root when they don't find the requested url.
The Jigsaw's configurations lets me do this only if I disable pretty-url in all site.

Other scenario is to create a file that I want to use as php but output as .xpto.

damiani:
This is only an issue when using `.md` files, because of the way Jigsaw builds pages from markdown files. You can accomplish a URL ending with `.html` at the moment by using a `blade.php` file instead of a `blade.md` file:

> 404.html.blade.php

```php
---
extends: _layouts.master
title: "404 Error"
---
@section('body')
    <h1>Oops! Maybe you did lost your path.</h1>
@endsection
```

If you wanted to still use markdown, rather than HTML, for formatting the page content, you could `@include()` the markdown content from a separate file:

```
@section('body')
    @include('_404.md')
@endsection
```

We're working on a fix that allows `.md` filenames to contain internal `.`s, which would solve this issue without having to add support for the additional `permalink` key.

PR #95, opened 2017-04-22
Add ability to register custom blade tags

TheNodi:
See #67.

To register a blade directive:
```php
<php
// blade.php

return [
    'datetime' => function ($expression) {
      return '<?php echo date("Y-m-d H:i:s", ' . $expression . '); ?>';
    }
];
```

I didn't add the $page parameter because it was calculated using get_defined_vars() on the resolved blade instance.

### DIFF
--- a/jigsaw
+++ b/jigsaw
@@ -14,6 +14,7 @@ use Mni\FrontYAML\Bridge\Symfony\SymfonyYAMLParser;
 use Mni\FrontYAML\Markdown\MarkdownParser;
 use Mni\FrontYAML\Parser;
 use Mni\FrontYAML\YAML\YAMLParser;
+use TightenCo\Jigsaw\BladeFile;
 use TightenCo\Jigsaw\CollectionDataLoader;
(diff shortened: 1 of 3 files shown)


damiani:
Excellent! I think we should change the name from `BladeFile` to `BladeDirectivesFile` so it doesn't get confused with the `BladeHandler`. After that change, it's good to go!

TheNodi:
Sure, done!

damiani:
Thanks for this contribution, @TheNodi! 👍

damiani:
Yes, this was added before component and include aliasing was added to Laravel, but it would be worthwhile supporting this. For now, though, I don't want to introduce a breaking change for anyone using the current method of registering directives. We can expose the `BladeCompiler` to `blade.php` easily, though, by passing it to the constructor of `BladeDirectivesFile` and assigning it to `$bladeCompiler`, rather than injecting it in the `register` method. I'd welcome a PR with that approach!

JoshMoreno:
Yea agree on not introducing breaking changes. Too quick.

Here ya go #204

PR #96, opened 2017-04-23
Copy files without an extension properly

paulredmond:
When I tried to publish my Jigsaw site with Github Pages, I noticed that `jigsaw build production` copied the `CNAME` file over incorrectly from `source/` with a trailing `.` character.

screen shot 2017-04-23 at 12 43 04 am

I added a check for an empty `$type` and returns the path without the `.` in the file.


damiani:
👍

PR #97, opened 2017-04-23
Enforce SPR-2 syntax.

lucasmichot:
Enforce SPR-2 syntax.


damiani:
❤️️ Me-agrees

lucasmichot:
❤️️ Me-think so

PR #121, opened 2017-07-10
Pagination: Set paginator variable 

andrewvergel:
The paginator works for me when I replace this line!


damiani:
Can you elaborate on this change? Pagination is working fine in my testing, and this change breaks it. What issue were you running into?

Issue #99, opened 2017-04-24
Collection isn't working


imam:
Hello.

The collection isn't working at all. I've tried to add `path` to `posts` in the `collection` array, not working. But the `jigsaw console` do notice the `_posts` folder because it would throw me an error if it isn't there, but it just won't render any item in the folder. What am I did wrong?

![2017-04-24_114849](https://cloud.githubusercontent.com/assets/8213031/25323148/a5d3c374-28e7-11e7-939b-dc4068d6bdeb.png)

The picture note:
1. The left editor is the `1-helloworld.md`
2. The right editor is the `config.php`

damiani:
I can't see any issues with your code (although the file `1-helloworld.md` is set to extend the `master` template, so your `post` template won't be used), though it's possible there's some issue relating to running on Windows. Do you get any error message when trying to build the site? 

If you upload a zip of your site, or link to a repo, I can test it out more fully.

imam:
There aren't any error, it just won't render the output.

For the repo: https://github.com/imam/imamtech-jigsaw

imam:
![2017-04-25_094647](https://cloud.githubusercontent.com/assets/8213031/25366791/39091144-299c-11e7-8ca2-57934999bb50.png)

damiani:
Your code builds fine for me when I run it on my mac, but something is causing the collection build to fail silently when I test it on Windows. I'm not certain yet of the cause.

damiani:
Actually, scratch that, it's working fine for me on Windows. At least, on Windows running the Ubuntu bash, which it _looks_ like you are running as well? (In my earlier test, I had mistakenly cleared out `config.php` so it had appeared to not work.) So I'm a bit stuck, since I can't reproduce the error when building from your repo.

![imam-test](https://cloud.githubusercontent.com/assets/357312/25564323/da418df8-2d7e-11e7-9714-7503e0934725.png)

VictorAvelar:
@imam @damiani Just hit the same problem, everything works fine but it does not generate the output files, the data is parsed because I can loop through the collection as $posts.

Do you manage to solve this?

damiani:
I wasn't able to reproduce the issue. Could you post a repo of your code, as well as some details about your setup? Running on Windows, using the Ubuntu bash shell, @imam's code built fine for me. I'd love to give yours a try to see if I can recreate the problem.

VictorAvelar:
[...]

PR #126, opened 2017-07-21
The actual fix for collections on windows system

garhbod:
The problem comes from windows paths using the `\` rather than `/`
Altered topLevelDirectory and relativePath to handle both `\` or `/`

### DIFF
--- a/src/File/InputFile.php
+++ b/src/File/InputFile.php
@@ -13,14 +13,16 @@ class InputFile
 
     public function topLevelDirectory()
     {
-        $parts = preg_split('/[\\/\\\]+/', $this->relativePath());
+        $parts = explode(DIRECTORY_SEPARATOR, $this->relativePath());
 
         return count($parts) == 1 ? '' : $parts[0];
     }
 
     public function relativePath()
     {
-        return trim(str_replace($this->basePath, '', $this->file->getPathname()), "\\/");
+        $relative_path = str_replace(realpath($this->basePath), '', realpath($this->file->getPathname()));
+
+        return trim($relative_path, DIRECTORY_SEPARATOR);
     }
 
     public function bladeViewPath()
--- a/src/File/InputFile.php
+++ b/src/File/InputFile.php
@@ -13,14 +13,14 @@ class InputFile
 
     public function topLevelDirectory()
     {
-        $parts = explode('/', $this->relativePath());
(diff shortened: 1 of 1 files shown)


quickliketurtle:
I think a cleaner solution would be to just use `DIRECTORY_SEPARATOR` instead of the hard coded '/'.

cmosguy:
@damiani I have confirmed that this patch fixes the problem of building the posts pages for a collection.  I was running in to the same error.  Can you please merge in ASAP, that would be super awesome!

Thanks @garhbod for the fix!

src/File/InputFile.php
```diff
@@ -13,14 +13,14 @@ public function __construct($file, $basePath)
 
     public function topLevelDirectory()
     {
-        $parts = explode('/', $this->relativePath());
+        $parts = preg_split('/[\\/\\\]+/', $this->relativePath());
```
damiani: I agree with @quickliketurtle that this would be better:
```
$parts = explode(DIRECTORY_SEPARATOR, $this->relativePath());
```

damiani:
Looks good, though I agree it would be better to use DIRECTORY_SEPARATOR in place of `preg_replace`. Could you make that change, and then if @cmosguy and @garhbod can confirm that it still solves their Windows issues, we can merge this in!

garhbod:
I fail to see how it is better? The regex handles both types of slashes and also multiple slashes `//` in the path. Covers all bases

damiani:
Easier to read than the regex, and gives more of an explicit indication of why it's there in lieu of the original `explode('/'...)`. Why do we need to worry about multiple slashes?

garhbod:
DIRECTORY_SEPARATOR has been implemented and tested. Needed to add realpath() to relativePath method to convert all the paths to the correct directory seperator for the entire path

damiani:
Fantastic, I'm testing this now. If it works for @adrianR84 and @cmosguy from over on https://github.com/tightenco/jigsaw/issues/99 then we'll be good to go!

garhbod:
@damiani Are pages themselves added to a collection at all?
I'm wanting to make a dynamic menu but $pages isn't set at all

damiani:
@garhbod Jigsaw doesn't set a `$pages` variable, only `$page` for the current page, and variables for whatever collections you've defined (e.g. `$posts`). At the moment, there isn't a way to build a menu dynamically for items that aren't contained within a collection.

Issue #128, opened 2017-07-27
[Proposal] Add ability to house multiple sites in one installation


adamkelso-rg:
Before going further, I'll specify that we're currently building out this functionality for a few things we need at work. This proposal is to see if Tighten would be interested in a pull request with the results when it's done. Thanks so much for releasing Jigsaw as open source.

## Overview
It would be nice to house multiple sites in one Jigsaw installation, with shared resources and views. This would allow several static sites to share resources, like templates or themes, and still have separate content and skins.

From the command line a user could then choose to build any specific site or all of them. This would be extremely helpful for occasions when a resource shared by multiple sites changes and everything needs to be updated.

## Our Current Plan
The idea so far has been to include a _config.php_ file in the root of the installation, specifying the source locations and build destination of each **domain**. During the build process, the build job iterates over each domain and simply loads it's specific config file and then builds like normal.

Also in the root config is a **namespaces** array that specifies source locations of views that can be turned into Laravel view modules. By doing so, any domain can use the namespaced resources with the syntax `@include('Module::folder.view')`.

Thanks again for the work on Jigsaw and the other Laravel community efforts.

damiani:
Yes, this would be a great addition. My initial thoughts are to keep the definition in the root `config.php` as simple as possible, so perhaps just a `sites` array with keys for each site that point to the `source` locations, and then let each individual site specify their build locations in their own `config.php` files. something like:

```php
return [
    'sites' => [
        'site-1' => 'source_main',
        'site-2' => 'source_secondary'
    ],
    'namespaces' => [
        'Shared' => 'shared_resources',
    ],
];
```

...then, you would run `jigsaw build site-1` for a single site, or just `jigsaw build` to build all of them.

I look forward to your PR!

adamkelso-rg:
Great!

That's pretty close to what I have so far, except I've been editing with the build destination in the root config.

```php
return [
    'domains' => [
        'directory-of-site-1' => 'build-destination-of-site-1',
        'directory-of-site-2' => 'build-destination-of-site-2'
    ],
    'namespaces' => [
        'parent' => 'parent-directory',
        'another' => 'somewhere-else'
    ]
];
```

However, I think you're right, it would make more sense to leave that in each site's configuration.

Also, I'm currently assuming the directories for each site have a **source** directory inside, and a **config.php** at the root of each site. This sound like a fair assumption to you, or do you think the exact source and exact config file should be specified in the sites array?

[...]

PR #130, opened 2017-07-28
Add Multiple Site Support

adamkelso-rg:
This PR is in response to issue #128 and adds the following conventions to support multiple sites.

1. If the installation should support multiple sites, the **config.php** in the root directory should contain a "sites" array with the keys being the short name of the site, and the values being the site location.

```php
// in /config.php

"sites" => [
    'site-1' => 'site-1',
    'site-2' => 'some-directory/site-2'
]
```

2. The directory destination each site points to should include a **source** directory, and a **config.php** file specific to that site. Each site's config file can follow the normal Jigsaw standards and also specify a build destination. HOWEVER, be aware that all build destinations are in relation to the root directory because the current working directory is not changed.
```php
// Example directory structure
/config.php
/jigsaw
/site-1
    /config.php
    /source
        /index.php
        /css
        /js
/some-directory
    /site-2
        /config.php
        /source
            /index.php
            /css
            /js
```

3. In the root **config.php** a new "modules" directory can specify the locations of view modules. These follow the standard Laravel convention that allows blade files to use a short `::` syntax to specify resource locations.

```php
// In /config.php
"modules" => [
    'First' => 'path/to/first/module',
    'Second' => 'path/to/second/module'
]

// Somewhere in a view
@extends('First::_layouts.master')

// or
@include('Second::_partials.navigation')
```

4. In order to maintain backwards compatibility, the following conventions have been followed with terminal commands:
    - If no "sites" index exists in the root level **config.php** file, all commands act exactly as they have. If a "modules" index exists, modules are added whether or not a sites index exists.
    - If a "sites" index exists, the `build` command iterates through each site and builds each serially.
    - A specific site can be specified with the `--site="sitename"` option or it's `-s sitename` alias. Doing so causes only the specified site to build. I decided to make this an option rather than a required argument because until now the `env` has been the argument immediately following the `build` command.
    - If the `serve` command is used and a "sites" index exists in the main config, the first site in the array will be attempted.
    - A site can be specified for the `serve` command by specifying the `--site="sitename"` argument or `-s sitename`.
    - No matter which site is served, if the site has not been built, the standard warning is shown before the command exits.

### Further Ideas
Currently there is no way to share "collections" between sites. It would be nice to be able to specify a location in any one collection's config array so that in the same way this PR adds the ability to share resources, collections could share content. If this sounds like an interesting idea, I'll be building that for our own stuff here and can make another PR.

### DIFF
--- a/jigsaw
+++ b/jigsaw
@@ -103,11 +103,11 @@ $container->bind(Factory::class, function ($c) use ($cachePath) {
 
     $finder = new FileViewFinder(new Filesystem, [$c['buildPath']['source']]);
 
-	  if(isset($c['config']['modules'])) {
-		    foreach($c['config']['modules'] as $namespace => $path) {
-			      $finder->addNamespace($namespace, $path);
-		    }
-	  }
+    if(isset($c['config']['modules'])) {
+        foreach($c['config']['modules'] as $namespace => $path) {
+            $finder->addNamespace($namespace, $path);
+        }
+    }
 
     return new Factory($resolver, $finder, Mockery::mock(Dispatcher::class)->shouldIgnoreMissing());
 });
--- a/jigsaw
+++ b/jigsaw
@@ -103,11 +103,11 @@ $container->bind(Factory::class, function ($c) use ($cachePath) {
 
     $finder = new FileViewFinder(new Filesystem, [$c['buildPath']['source']]);
 
-	if(isset($c['config']['modules'])) {
-		foreach($c['config']['modules'] as $namespace => $path) {
-			$finder->addNamespace($namespace, $path);
-		}
-	}
+	  if(isset($c['config']['modules'])) {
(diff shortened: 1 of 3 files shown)


adamkelso-rg:
@damiani Just an update on this one. I just submitted a change that fixes the occasion where a build destination is not specified on a specific site. This was causing sites to overwrite the build_local directory, so only the last site was shown. Now, it's set to build a site to build_local/site_name only if there are multiple sites and no build destination is specified in a specific site's config.

damiani:
My thinking now is that we've gotten much of this capability through the ability to specify view hint paths (see #393 and [the docs](https://jigsaw.tighten.co/docs/content-blade/#specifying-blade-hint-paths)), so multiple sites could at least share templates; and that the rest of the features here would add a lot of complexity for something that I feel would have pretty rare usage.

PR #118, opened 2017-07-02
Proof of concept: remote collections

DRogueRonin:
Proof of concept as suggested here: https://github.com/tightenco/jigsaw/issues/113
With my ideas from here: https://github.com/tightenco/jigsaw/issues/117

**config.php**
```
<?php

return [
    'baseUrl' => '',
    'production' => false,
    'collections' => [
        'posts' => [
            'remote' => function() {
                $posts[] = json_decode(file_get_contents('https://pokeapi.co/api/v2/pokemon/1/'));
                $posts[] = json_decode(file_get_contents('https://pokeapi.co/api/v2/pokemon/4/'));
                $posts[] = json_decode(file_get_contents('https://pokeapi.co/api/v2/pokemon/7/'));
                
                return array_map(function($v) {
                    return [
                        'extends' => '_layouts.master',
                        'section' => 'body',
                        'filename' => $v->name,
                        'title' => ucfirst($v->name),
                        'category' => $v->types[0]->type->name,
                        'date' => date('Y-m-d'),
                        'content' => "# {$v->name}\n<img src='{$v->sprites->front_default}'>", # refer to layout instead
                    ];
                }, $posts);
            },
            'path' => 'blog/{date|Y}/{filename}',
        ],
    ],
];
```

PR #136, opened 2017-08-07
Dispatcher Event/Listener after build completed

cmosguy:
Hey @damiani ,

I think I figured out a reasonable solution.  The idea here is we can implement Event classes based off the `Illuminate\Dispatcher`.  Then in the uses `bootstrap.php` they can hack in listeners as they so choose.

For example, my `bootstrap.php` for adding to Algolia is here:

```php
use Illuminate\Events\Dispatcher;
use TightenCo\Jigsaw\DataLoader;
use TightenCo\Jigsaw\Events\BuildHasCompleted;


$container->bind(Dispatcher::class, function ($c) {

    $dispatcher = new Dispatcher();

    $dispatcher->listen([BuildHasCompleted::class], function ($event) {
        $dotenv = new Dotenv\Dotenv(__DIR__);
        $dotenv->load();


        $client = new \AlgoliaSearch\Client(getenv('ALGOLIA_APP_ID'), getenv('ALGOLIA_ADMIN_API_KEY'));

        $index = $client->initIndex('prod_html');

        /** @var \TightenCo\Jigsaw\DataLoader $dataLoader */
        $dataLoader = $event->app->make(DataLoader::class);

        /** @var TightenCo\Jigsaw\Collection\Collection $siteData */
        $siteData = $dataLoader->load($event->app->buildPath['source'], $event->app->config);

        $siteDataArr = $siteData->toArray();

        $val = $index->addObjects($siteDataArr);

    });

    return $dispatcher;
});

```

Now, I know this is not the most elegant solution within the `bootstrap.php`.  I think the end goal here is to put the listeners into the `config.php` instead.  In fact, I still need to implement logic here that will only execute when the environment is  `production`.  In my example above, ignore the DotEnv stuff, I figured out that I can create a separate `config.production.php` for the Algolia API and SECRET keys. 

What I need to figure out how to do is wrap all this into a class.  Currently, this would mean i need to make a new composer.json entry that takes in psr-4 of the `source` directory and I put Listeners in there?  What would you recommend?

Cheers!
Adam

### DIFF
--- a/src/Console/BuildCommand.php
+++ b/src/Console/BuildCommand.php
@@ -37,8 +37,8 @@ class BuildCommand extends Command
 
         $this->app->make(Jigsaw::class)->build($env);
         /** @var $dispatcher Dispatcher */
-        $dispatcher = $this->app->make(Dispatcher::class);
-        $dispatcher->fire(new BuildHasCompleted(Jigsaw::class));
+//        $dispatcher = $this->app->make(Dispatcher::class);
+//        $dispatcher->fire(new BuildHasCompleted(Jigsaw::class));
         $this->info('Site built successfully!');
     }
 
--- a/src/Console/BuildCommand.php
+++ b/src/Console/BuildCommand.php
@@ -37,8 +37,8 @@ class BuildCommand extends Command
 
         $this->app->make(Jigsaw::class)->build($env);
         /** @var $dispatcher Dispatcher */
-//        $dispatcher = $this->app->make(Dispatcher::class);
-//        $dispatcher->fire(new BuildHasCompleted(Jigsaw::class));
+        $dispatcher = $this->app->make(Dispatcher::class);
+        $dispatcher->fire(new BuildHasCompleted(Jigsaw::class));
         $this->info('Site built successfully!');
     }
 
--- a/src/Console/BuildCommand.php
+++ b/src/Console/BuildCommand.php
@@ -38,7 +38,7 @@ class BuildCommand extends Command
         $this->app->make(Jigsaw::class)->build($env);
         /** @var $dispatcher Dispatcher */
         $dispatcher = $this->app->make(Dispatcher::class);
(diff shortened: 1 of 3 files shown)


damiani:
I definitely like the idea of firing a handful of events at various points in the build process for people to listen for—perhaps before the build starts, before and after each file is processed, and after the entire build completes. I think we need to come up with a few good real-world examples of how these various event hooks would be used, so we can tailor the solution.

pfeiferbit:
I could give you a real world example of how I would use it:

For every build process I want to consume the API of a Mantis bugtracking system to retrieve data for given bug IDs. I then want to use that piece of data like a collection of Markdown files in a folder, rendering each with a given template.

damiani:
@pfeiferbit That functionality will be covered in https://github.com/tightenco/jigsaw/pull/118, which is under development at the moment.

damiani:
@cmosguy I've combined your idea with a couple others into a new PR (https://github.com/tightenco/jigsaw/pull/189), take a look and let me know your thoughts there. With that feature, you can hook into events in `bootstrap.php` by adding them as closures or, for more complex code, separate classes. In your Algolia example, you would be able to add this code to `bootstrap.php` to index all your collections after they've been built:

```php
use AlgoliaSearch\Client;

$events->afterBuild(function ($jigsaw) {
    $objects = $jigsaw->getCollections()->flatten(1)->map(function ($item) {
        return [
            'title' => $item->title,
            'content' => $item->getContent(),
            'url' => $item->getUrl(),
        ];
    });

    (new Client($jigsaw->getConfig('algolia_app_id'), $jigsaw->getConfig('algolia_api_key')))
        ->initIndex('prod_html')
        ->addObjects($objects);
});
```

The `$jigsaw` variable (which is an instance of `TightenCo\Jigsaw`) gives the event listener access a bunch of new helper methods, so that getting/setting site information and reading/writing files is simpler for the user, without having to muck about with the internals of `DataLoader`.

Issue #142, opened 2017-08-30
Add compatibilty with Laravel 5.5


michellaurent:
There is a breaking change in illuminate/view 5.5.

`Interface 'Illuminate\View\Engines\EngineInterface' not found`

https://github.com/illuminate/view/commit/4c38e3b6ba6325bb1a23a379cc32e9751ec5b536

`use Illuminate\View\Engines\EngineInterface;` should be replaced with
`use Illuminate\Contracts\View\Engine as EngineInterface;` in 
* TightenCo\Jigsaw\View\BladeMarkdownEngine
* TightenCo\Jigsaw\View\MarkdownEngine

PR #143, opened 2017-08-30
Fix https://github.com/tightenco/jigsaw/issues/142

hyperized:
Fixes https://github.com/tightenco/jigsaw/issues/142

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,9 +23,9 @@
         }
     },
     "require": {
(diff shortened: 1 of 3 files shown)


damiani:
Looks good...could you update the versions of the Illuminate dependencies in `composer.json` to `^5.5`? Then this should be good to go.

hyperized:
@damiani done :)

PR #144, opened 2017-08-31
Escape blade braces as well

mattstauffer:
Using tokens and `<?=` (we could replace that with `<?php echo` i guess if we want) to avoid the braces of one replacement getting replaced again by the next.

It's *possible* it's overkill to use BOTH php echo to avoid this problem AND merging the PHP open tag issue with the blade tag replacement issue like I did here. I'm open the the suggestion of trying to   separate the PHP replacement back out to its own method if we wanted.

It looks like I may have broken some sort of newline thing here. @damiani can you suss out what I broke there from looking at the snapshots?

### DIFF
--- a/src/Handlers/MarkdownHandler.php
+++ b/src/Handlers/MarkdownHandler.php
@@ -75,23 +75,21 @@ class MarkdownHandler
 
     private function getEscapedMarkdownContent($file)
     {
-        $markdownFileContent = $this->escapePhpOpenTags($file->getContents());
+        $content = str_replace('<?php', "__PHP_OPEN__", $file->getContents());
 
         if ($file->getFullExtension() == 'md') {
-            $markdownFileContent = $this->escapeAtSymbolInMarkdown($markdownFileContent);
+            $content = str_replace(
+                ['{{', '}}', '{!!', '!!}', '@'],
+                ['__OPEN_CURLIES__', '__CLOSE_CURLIES__', '__OPEN_DANGER__', '__CLOSE_DANGER__', '__LONE_AT__'],
+                $content
+            );
         }
 
-        return $markdownFileContent;
-    }
-
-    private function escapePhpOpenTags($content)
-    {
-        return str_replace("<?php", "<{{'?php'}}", $content);
-    }
-
-    private function escapeAtSymbolInMarkdown($content)
-    {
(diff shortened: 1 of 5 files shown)


src/Handlers/MarkdownHandler.php
```diff
@@ -75,23 +75,21 @@ function ($path) use ($pageData, $layout) {
...
-    }
-
-    private function escapeAtSymbolInMarkdown($content)
-    {
-        return str_replace("@", "{{'@'}}", $content);
+        return str_replace(
+            ['__OPEN_CURLIES__', '__CLOSE_CURLIES__', '__OPEN_DANGER__', '__CLOSE_DANGER__', '__LONE_AT__', '__PHP_OPEN__'],
+            ["<?= '{{' ?>", "<?= '}}' ?>", "<?= '{!!' ?>", "<?= '!!}' ?>", "<?= '@' ?>", "<?= '<?php' ?>"],
+            $content
+        );
```
damiani: I think this would be a little cleaner:

```php
private function getEscapedMarkdownContent($file)
{
    $replacements = ["<?php" => "<{{'?php'}}"];

    if ($file->getFullExtension() == 'md') {
        $replacements = array_merge([
            "@" => "{{'@'}}",
            "{{" => "@{{",
            "{!!" => "@{!!",
        ], $replacements);
    }

    return strtr($file->getContents(), $replacements);
}
```

- we don't need to worry about closing `}}` or `!!}`, since Blade won't care about those if the opening braces aren't present
- using `strtr` allows us to avoid the intermediate `__TOKEN__` dance, since `strtr` won't process a value twice
- since the goal is to set the markdown up to be correctly processed by Blade, I feel like using Blade's `{{` for echoing and `@` for escaping fits better than resorting to `<?= ... ?>`

As for the newline issue, it seems that using Blade to replace the PHP open tag via `<{{'?php'}}` handles newlines correctly, whereas replacing it with `<?= '<?php' ?>` squashes the newline if present. I didn't dig into why, but I figure if the Blade method is parsing it correctly, we might as well stick with that.

mattstauffer: 👌🏻 I'll pull it down and test it. Thanks!

mattstauffer:
@damiani updated and ready for review again (it's 100% your code now lol)

Issue #154, opened 2017-09-14
[Help] Trouble with multi-variable collection sorting


dgursh:
I have the following variables in my YAML files for Lessons.

```
Lessons.md
---
subject: "php"
course: "basics"
courseName: "php-basics"
title: "PHP Intro"
description: "this is a description..."
orderNumber: 1
---
```
My goal is to sort the lessons first by "courseName" and then by the "orderNumber". That way I can progress through all of the lessons of the course in order.

In my config I have the following for the sorting of lessons collection
```
'lessons' => [
            'sort' => ['orderNumber',  'courseName']
]
```

The result is that it appears to be taking into account orderNumber and not the courseName. The following two lessons appear next to each other even though I'm also sorting based on courseName.

```
php-intro.md
---
subject: "php"
course: "basics"
courseName: "php-basics"
title: "PHP Intro"
description: "this is a description..."
orderNumber: 1
---

CSS-intro.md
---
subject: "css"
course: "basics"
courseName: "css-basics"
title: "CSS Intro"
description: "this is a description..."
orderNumber: 1
---
```

I even tried reversing the order, like so. But that gives an even crazier order that is not correct.
```
'lessons' => [
            'sort' => ['courseName',  'orderNumber']
]
```

Any advice on how to fix / debug?

damiani:
It should work the first way you have it, with `'sort' => ['orderNumber',  'courseName']`, but it seems to be failing when there are multiple `orderNumbers`. I'll take a closer look and work on a fix.

dgursh:
Thank you. I recreated all of the lesson markdown files and it was working great 
```
'sort' => ['courseName',  'orderNumber']
```
but only when I had duplicates of a single orderNumber (5 lessons with orderNumber of 1). But once, I started adding more lessons with duplicate orderNumber's of 2, the order fell apart.

dgursh:
Ok, thank you. Let me know when you've figured out a fix. Would you like me to upload a sample repo project with the code for you to look at? @damiani

dgursh:
Hi @damiani  I set up a sample repo that tests this layout

https://github.com/dgursh/bug

You'll notice when you navigate to the home page the the lesson links are out of order, despite the sorting variables that I use. Let me know your thoughts.

dgursh:
@damiani where in the src code is the custom sorting functions? I can take a look at it as well.

PR #155, opened 2017-09-15
Fix sorting of collection items by multiple variables

damiani:
Fixes https://github.com/tightenco/jigsaw/issues/154

When working with collections, the `sort` parameter in `config.php` can accept an array of multiple values. This was previously done with chained calls to the Illuminate/Collection `sortBy` or `sortByDesc` methods, but that doesn't do what we expect; instead (as referenced in https://github.com/laravel/internals/issues/11), it does not maintain the order of any of the preliminary sorts, and in the end only sorts by the final `sortBy` clause.

This has been fixed by using the `sort` method, which calls PHP's `usort` under the hood and allows us to do our own comparison. I've used what I believe is a pretty elegant solution from https://stackoverflow.com/a/27364612/4043861, which makes the comparison between adjacent items by stepping through each criterion (each variable, in our case) and tests them with `strcmp()`, returning the first non-zero value from a nested ternary expression:

```php
$items->sort(
    function ($a, $b) {
        // sort by column1 first, then 2, and so on
        return strcmp($a->column1, $b->column1)
            ?: strcmp($a->column2, $b->column2)
            ?: strcmp($a->column3, $b->column3);
    }
);
```

### DIFF
--- a/src/Collection/Collection.php
+++ b/src/Collection/Collection.php
@@ -68,19 +68,19 @@ class Collection extends BaseCollection
     private function compareItems($item_1, $item_2, $sortSettings)
     {
         foreach ($sortSettings as $setting) {
-            $comparison = $setting['direction'] * strcasecmp(
-                $this->getValueForSorting($item_1, array_get($setting, 'key')),
-                $this->getValueForSorting($item_2, array_get($setting, 'key'))
-            );
+            $value_1 = $this->getValueForSorting($item_1, array_get($setting, 'key'));
+            $value_2 = $this->getValueForSorting($item_2, array_get($setting, 'key'));
 
-            if ($comparison) {
-                return $comparison;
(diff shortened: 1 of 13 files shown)


dgursh:
@damiani I applied the new code you wrote to my collection.php file. The order is better, however, in my specific use-case of the orderNumber. It is placing the following lessons orderNumber: 11 next to lessons with orderNumber: 1 ... (orderNumber 12 and orderNumber 2). Not sure why that is. It's like it is only taking into account the the first number in a two digit number...

damiani:
Yes, I'm working on that.

damiani:
@dgursh Numeric values are sorted correctly now, and your test repo checks out for me.

dgursh:
@damiani works great. I had a feeling the issue was with the strcasecmp function.

Issue #145, opened 2017-09-08
Registration and Trademark symbols in variables are rendered in final file name


standingdreams:
I recently upgraded Jigsaw to include the pull request #141. I needed a way to include an `.htaccess` for site redirects. I am using Blade/Markdown files to render pages from a parent template for products and accessories on a site. The name of the product is being used as the file name. 

Before upgrading trademark (™) and registration (®) symbols were being stripped out of the final file name. However, after upgrading it is now rendering ™ and ® as "tm" and "r", respectively. I was able to rectify the issue (of course after the client noticed the problem). I simply had to create another variable for the file name that did not include ™ or ® in it. 

I wanted to bring it up as an issue, just in case anyone else comes across this problem.

dgursh:
Why are you using those symbols in the file name? If you're worried about copyright, there's really nothing you can do since it is a static website. You're best bet is to put copyright notice in the footer of your website and include a terms of service and privacy policy

standingdreams:
@dgursh The goal was not to have those symbols as apart of the file name at all. I totally can see what I left off in this explaining this issue: the title variable was being used to define the path in the collection.

Example
`'path' => 'blog/{title}'   // 'blog/Title Of First Post'`

Previously the title could consist of trademark (™) and registration (®) symbols and they would be stripped from the path. However, after upgrading Jigsaw to include the pull request #141, the path began to render ™ and ® as "tm" and "r", respectively.

dgursh:
Ok, so you are including trademark and registration symbols in the URL paths of your pages. Out of curiosity, why are you doing that? I don't think you can include html entities ™ and ® in a URL without first encoding it.

standingdreams:
No, the trademark and registration symbols should not be apart of the URL paths at all. However, the `title` variable that was used in the templates had them in there. When creating a collection, you can use a variable from the a Markdown/Blade file as the path. I was using "Product 1®" to show as the title within the template but also using it for the path. 

Initially, this worked fine. It seems the master branch breaks that functionality.

dgursh:
I understand what you are saying. But the path route defined in your config file is exactly that... the path (the URL). So, if you are using a title variable that includes those characters it will be included in the URL.

standingdreams:
Previously, that wasn't the case. I had trademarks and registration symbols all in that title variable. It would only output the characters that were not symbols. However, after upgrading Jigsaw, that functionality was broken.

[...]

PR #156, opened 2017-09-15
Strip URL-invalid characters from path names

damiani:
Resolves https://github.com/tightenco/jigsaw/issues/145#issuecomment-329870894

When specifying a variable to use as a path name (e.g. `title`) and _not_ slugifying it (as with `'path' => '{title}'`), we need to strip out any characters that are not valid in URLs. This was the behavior in v1.0, but was inadvertently changed since then. I'm opting not to use `urlencode` because we want to maintain `/` characters (for nested paths and date processing), nor FILTER_VALIDATE_URL, which would remove spaces. A good old-fashioned regex restores behavior that is consistent with v1.0.

### DIFF
--- a/src/PathResolvers/CollectionPathResolver.php
+++ b/src/PathResolvers/CollectionPathResolver.php
@@ -89,7 +89,7 @@ class CollectionPathResolver
             $param = ltrim($param, $param[0]);
         }
 
-        $value = array_get($data, $param, $data->_meta->get($param));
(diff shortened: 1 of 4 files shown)


dgursh:
@damiani At that point, don't you think we should automatically just be calling str_slug, laravel helper on every variable that is used in the path config?

damiani:
No, I think there are plenty of valid cases were people will want to use the unaltered variable contents. Adding `-` to the `path` (e.g. `{-title}`) will slugify, if desired.

mattstauffer:
👌

PR #164, opened 2017-10-07
Command to generate `.md` and `.blade.md` files in a fly

wilburpowery:
This PR adds a new `make:post` command. This command generates a file with some starting template. I would like to add some other stuff to it but wanted to check what other customizations you guys had in mind. 🤠

### DIFF
--- a/jigsaw
+++ b/jigsaw
@@ -3,12 +3,14 @@
 
 use TightenCo\Jigsaw\Console\BuildCommand;
 use TightenCo\Jigsaw\Console\InitCommand;
+use TightenCo\Jigsaw\Console\MakePostCommand;
 use TightenCo\Jigsaw\Console\ServeCommand;
 
 require_once(__DIR__ . '/jigsaw-core.php');
 
 $app = new Symfony\Component\Console\Application('Jigsaw', '1.0.8');
 $app->add($container[InitCommand::class]);
+$app->add($container[MakePostCommand::class]);
 $app->add(new BuildCommand($container));
 $app->add(new ServeCommand($container));
 $app->run();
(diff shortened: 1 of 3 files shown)


damiani:
This is a good start, though I think we should think about building something that's more generalizable, rather than adding a command for something specific like a `post`. 

I'm envisioning something that would allow the user to define their own template stubs (like your `/Console/stubs/markdown-file.md`, but user-defined)—where they could include YAML header variables with default values, as well as body content for the template. There might also be a couple places where we would want to involve some logic—setting a `date` variable to the current date, for example. So if someone builds a stub called `post.md`, for instance, they could call `jigsaw make post --title="My New Post"`, but they could also do the same for other content types.

The other issue I foresee is specifying where the newly generated file should end up; it's cumbersome to ask the user to specify the location each time, which makes me think maybe we should either link it to a collection via the `config` file, or by name (so `/templates/post.md` would build to the `_posts` directory automatically), or locate the stub (called `_template.md` or something similar) in the directory where the built file should end up.

I would love to hear your thoughts on ways we can implement this in a more generalized way.

wilburpowery:
I agree with all you said. Being a jigsaw user myself I'd love the idea of simply creating `/templates/post.md` stub and after running the make:`something-more-general` command, have it generate the file inside `/_posts`. I'll look into this more over the next few days and see what I can come up with. If you come up with anything for this feature hit me up on telegram [here](https://t.me/wilburpowery). 🤠

calebporzio:
Hey @wilburpowery - We just opened up a new WIP PR (#194) that outlines a new version of this proposal - thanks for the hard work putting this together. Right, now this PR is too tightly coupled to the "posts" collection - we want to open it up a bit.

We will personally be working on the expanded feature, but feel free to take a look and pitch in if you are so inclined.

Closing this for now.

wilburpowery:
Awesome @calebporzio, nice to see this feature picking up. I’ve been to busy to really keep working on this. Thanks for the heads up!

Issue #120, opened 2017-07-08
[Proposal] Plugins/Extensions for Jigsaw


DRogueRonin:
A way to to `composer require` a plugin would be nice, that for example introduces helper functions and partials I use for every project. So when updating a helper function or a general partial I don't have to go through every project and replace that, but just install the most recent version of my plugin and be done.

More detailed example: A Google Analytics extension you can require and just put your tracking code into `config.php` and place the function that outputs the code in the layout.

And/or maybe `npm install` a plugin to hook into the gulp build process.

cmosguy:
I would like to upvote this idea, does anyone have a recommended way of implementing this idea?  I want to build an Algolia indexer during the build process to hook into.

damiani:
Definitely open to this. Some things like the Google Analytics code can already be accomplished without any plugin need (just put your tracking code in `config.php` and include Google's snippet in your master template, echoing your code where appropriate), but a plugin/extension architecture would be useful, in addition to before/after middleware functionality for any custom pre-/post-processing that people might want to do.

damiani:
We should be able to piggy-back off of [this PR](https://github.com/tightenco/jigsaw/pull/189) to support this.

atymic:
@damiani @mattstauffer 

I would be keen to have a go at working on this. Would you be open to a PR with a WIP implementation?

damiani:
@atymic Yes, absolutely!

molerat619:
Any news on this? I want to import some plugins into my project like jQuery etc. 

Of course I can download the file manually and add it to the right folder, but I was wondering if you could `npm install jquery` and have Webpack actually using it in the build.

atymic:
@molerat619 this is completely different (this is back end/php plugins).

You should already be able to do that with laravel mix?

damiani:
@molerat619 Yes, that's already possible; install it via `npm` and adjust your `webpack.mix.js` file accordingly.

Issue #135, opened 2017-08-06
[Question] How to hook into an event in Console SiteBuilder command


cmosguy:
@damiani so I am trying to build an Algolia indexer that crawls through the markdown files that we use during the build process.  I am trying to understand how to hook into the SiteBuilder command.  I was reading this: http://symfony.com/doc/current/components/console/events.html but I could not figure out how to pull the EventDispatcher without pulling in the entire symfony/somfony package.

Is there an easy way to add an event listener that listens to `ConsoleEvents::TERMINATE` so that when the build process is completed, I can kick off another class that does the work I need to do.  For example, I am going to go through the data from the markdown content in my new class that I want to make.

I was trying to figure out how to get the Illumunate\Event to work, but I am struggling to figure out how to build the listener for this when the console event is fired and "TERMINATES".

Thanks!
Adam

damiani:
Presumably you would only want to run this when building for `production`, right? So we'd need to either hook it into the `config` files (i.e. via a middleware-type function that runs before or after the site files are processed, but conditionally based on environment) or, alternatively, if you went the Symfony-events route, you could check which options were provided to the `build` command.

There's no way I can think of doing this out-of-the-box, but I'm definitely open to exploring options for adding support for this. I just want to get a sense of some of the various use cases we can target, so we build something that's flexible and easy for users (such as those described in https://github.com/tightenco/jigsaw/issues/120).

cmosguy:
Thanks @damiani,

I created a pull request here: https://github.com/tightenco/jigsaw/pull/136

Thanks to the great @mattstauffer for inspiring us on his Containers education from his talks and books.  We can simply use the existing container in Jigsaw to make any Event/Listener we want.  Now, we just need to figure out how to get this more streamlined to add listeners and then update the Docs if you accept these ideas.


BTW, your middleware ideas are good.  Also, yes, I only want to build the index only for production.

Cheers!

PR #165, opened 2017-10-08
Foundation for extension development in Jigsaw

imam:
Referring to this issue https://github.com/tightenco/jigsaw/issues/120, I'm creating the foundation for Jigsaw's extension functionality development.

Here is my idea for this extension functionality development:

- The inspiration comes from Wordpress plugin development, that use `Action` and `Filter` name for their hooks. 

- Extension developers can do as easy as `$this->app->make(Action::class)->add('tag_to_be_hook_at', $callback)` in their class

- Users can install every Jigsaw's extensions that they want by adding to `extension` array in `config.php` the extension class.

It's a very raw idea, to duplicate Wordpress plugin functionality. Feedback is more than welcome 😄 

Note:
- Maybe we should create a single branch for extension functionality development
- I have to rename this `Remove Invalid Characters, from Paths?!'` directory so I can clone my fork to my computer


damiani:
@imam Thanks for the idea! Your approach seems similar to https://github.com/tightenco/jigsaw/pull/136/files (which arose out of Issue https://github.com/tightenco/jigsaw/issues/135), but using the "hooks" and "actions" nomenclature, instead of "events" and "listeners". Can you describe how your approach would be different from https://github.com/tightenco/jigsaw/pull/136? What tags would the actions hook into?

imam:
Hello @damiani, you're welcome!

Yes, I've read that pull request several minutes after I did create a pull request.

The only difference between mine and that is that I want to copy how Wordpress do their plugin development, maybe it will help lots of people that already familiar with that CMS when they want to use Jigsaw. Jekyll uses name `hook` rather than `listener` too. But their functionality is basically the same as how `listener` is used in `dispatcher`.

It's not that different. 

I'll look forward to which request will you pull and help the functionality development afterward :)

damiani:
I'm going to close this so we can keep development of this feature on a single PR, https://github.com/tightenco/jigsaw/pull/136.

PR #167, opened 2017-10-23
SEO tags support

darol100:
### Problem / Motivation
Most static website generator (at least the ones I have used in the past)  have the ability to add meta tags by default.

### Possible Solution
In this pull request, I added meta tags as site variables so it can be modified either by the blade template or by Markdown.

### DIFF
--- a/site/source/_layouts/master.blade.php
+++ b/site/source/_layouts/master.blade.php
@@ -1,10 +1,13 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="{{ $page->lang }}">
     <head>
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+        <meta name="description" content="{{ $page->description }}">
+        <meta name="keywords" content="{{ $page->keywords }}">
         <meta http-equiv="x-ua-compatible" content="ie=edge">
         <link rel="stylesheet" href="/css/main.css">
+        <title> {{ $page->title }}</title>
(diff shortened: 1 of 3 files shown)


site/source/_layouts/master.blade.php
```diff
@@ -1,10 +1,13 @@
...
-<html lang="en">
+<html lang="{{ $page->lang }}">
     <head>
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+        <meta name="description" content="{{ $page->description }}">
+        <meta name="keywords" content="{{ $page->keywords }}">
         <meta http-equiv="x-ua-compatible" content="ie=edge">
         <link rel="stylesheet" href="/css/main.css">
+        <title> {{ $page->title }}</title>
```
damiani: Remove extra space before `{{`

site/source/index.blade.php
```diff
@@ -1,5 +1,11 @@
 @extends('_layouts.master')
 
+{{-- Meta Data --}}
+@if ( $page->title = 'Page title' )@endif
+@if ( $page->description= 'Description' )@endif
+@if ( $page->keywords= 'Keywords, best' )@endif
+
+{{-- Content --}}
```
damiani: We can use YAML front matter for adding variables to Blade templates, so this can be simplified to:
```
---
title: Page Title
description: Page Description
keywords: Page, Keywords
---

@extends('_layouts.master')
...
```

PR #195, opened 2018-03-09
Add a confirmation prompt to the "init" command

calebporzio:
Right now, if you run `jigsaw init` on an existing jigsaw project, it will nuke your project by overwriting all the jigsaw files.

This PR should add a confirmation prompt to the command to alert users what they are about to do. It should be conditionally added based on if there is an existing `config.php` already or not.

image

### DIFF
--- a/src/Console/InitCommand.php
+++ b/src/Console/InitCommand.php
@@ -46,7 +46,7 @@ class InitCommand extends Command
             $this->info('Running it again may overwrite important files.');
             $this->info('');
 
-            if (! $this->confirm('Do you wish to continue?')) {
+            if (! $this->confirm('Do you wish to continue? ')) {
                 return;
             }
         }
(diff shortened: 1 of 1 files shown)


src/Console/InitCommand.php
```diff
@@ -32,9 +32,26 @@ protected function fire()
...
+    }
+
+    private function ifAlreadyScaffoldedWarnBeforeDoingTheFollowing($callback)
+    {
+        if ($this->files->exists($this->base . '/config.php')) {
+            $this->info('It looks like you\'ve already run "jigsaw init" on this project.');
+            $this->info('Running it again may overwrite important files.');
+            $this->info('');
+
+            if (! $this->confirm('Do you wish to continue?')) {
```
damiani: Can you add a space after `?` so the user input gets a little breathing room?

calebporzio: good call

damiani: Not like anybody needs this feature.

PR #200, opened 2018-03-18
Fix package.json for mix scaffold

JoshMoreno:
Use local webpack rather than global.

Fix NODE_ENV variables so assets get optimized properly during production builds.


damiani:
Thanks @JoshMoreno !

PR #201, opened 2018-03-18
Jigsaw node module integration

JoshMoreno:
I'd like to discuss before a decision is made to or not to merge.  And I'm open to transferring ownership of the node module to tightenco so it's all within the same family.

The current module can be found here - [https://github.com/JoshMoreno/node-jigsaw](https://github.com/JoshMoreno/node-jigsaw)

After using Jigsaw the past few weeks, I've really taken a liking to it.  There were just a few things I thought could be better.

### Problems

1. **CSS only changes don't get injected.**
It causes a full page reload. Not a big deal, but a bit of a drag when styling interactive components (i.e. menu only visible when triggered from a user click).
2. **`tasks` dir has quite a bit of hardcoded config**
This will make future upgrades more difficult for the user and possibly for maintainers to have to document/explain.
3. **Production webpack build isn't self contained**
It builds to source which muddies up the development commits.
4. **Build process requires two independent steps**
Currently a user has to have webpack compile production assets then have jigsaw do it's thing.

### Solutions
So not all of the above require a separate node module to solve, but I think it's where the solution belongs. And though I'm hesitant on separating this from the main package, I do think the benefits outweigh the cons.

1. Create our own watcher and only run a `jigsaw build $env` when we need to do so. That way the right files are getting changed at the right time.
2. Pull into the module and let the user override anything with a simple `Object.assign`. Also has the benefit of one less folder in the root. A small but very satisfying win in my book.
3. Can be solved with a `mix.inProduction()` check. But this needs to run after `jigsaw build production`.
4. The module essentially has the same `path()` function to check where jigsaw is installed. But this time it's installed in the `node_modules/.bin` so we can have a script that does `jigsaw build production && npm run production`. 

Hope all the above makes sense. 

**Any thoughts or concerns?**

---
### Give it a test run.

**composer.json**
```json
{
  "repositories": [
    {
      "type": "vcs",
      "url": "https://github.com/JoshMoreno/jigsaw"
    }
  ],
  "require": {
    "tightenco/jigsaw": "dev-jigsaw-node-module-integration"
  }
}

```
then..
```bash
composer install && npm install && npm run watch
```
Try changing some blade templates (reload), sass (inject), and js (reload).

Then to build production. Which should give you a `build_production` folder with optimized css and js. Source assets should be left unoptimized.

```bash
npm run jigsaw
```


damiani:
This looks very intriguing! I'll give it a test run...

damiani:
One important issue that I can foresee is that this breaks Mix versioning, i.e. you couldn't include the following in your master template:

```php
<link rel="stylesheet" href="{{ mix('css/style.css', 'assets/build') }}">
```

...since, as far as I can tell, you're essentially running `jigsaw build` to build the site _before_ assets are compiled, meaning no Mix manifest will exist when the `mix()` helper runs. This is why the current process includes building assets into `source` first, and then running `jigsaw build` to copy them to `build_local` or `build_production`.

JoshMoreno:
Ahhh. Yes. Good catch. So this really only affects number 3 right?

> 3. Production webpack build isn't self contained
> It builds to source which muddies up the development commits.

I'll need to good think about this one. 🤔

But in the meantime and perhaps moving forward I can revert the changes to solve number 3. I'll do some tests tomorrow to make sure `mix()` is working correctly.

JoshMoreno:
@damiani Yea I don't see a decent way of solving number 3. Every possible solution feels dirty and hacky. We definitely need to run webpack _before_ running  jigsaw. I've reverted the changes that attempted to solve this. 

Also added versioning, aka `mix()`, to the `master.blade.php` so it's more complete out of the box.

Any other concerns or things I can improve on here?

PR #220, opened 2018-05-26
Move from Sass to Less

mattstauffer:


GrimLink:
@mattstauffer why do you want this?
Most CSS frameworks are SCSS based.

Also, this would be better as an option.
Instead setting it hard to LESS.

mattstauffer:
@GrimLink Sass is heavier on the server, slower, and, while it originally had a lot more than Less does, no longer has enough to offer to make it worth it.

Anything is an option. Once you generate your app, you can just... change it. :)

incraigulous:
I want to officially vote against merging this in. SCSS is more popular now days, not to mention it's the default in Laravel.

damiani:
I'm gonna close this—at this point, I think we should probably remove the Sass default entirely, and ship with Tailwind instead :)

mattstauffer:
Amen!

incraigulous:
Amen!

PR #226, opened 2018-06-17
Performance & a Progress bar

abdumu:
I'm using Jigsaw for a project that has many source files, so I needed a short generation time to limit the down time (because I do the build in the production).

- Clean build folder after the generation, to make the down time short. (currently, cleaning happens before)
- Faster by 70%...
- & a Progress bar

Before
↓
After

To try,  copy the following to your composer.json:

```
{
    "repositories": [{
        "type": "vcs",
        "url": "https://github.com/abdumu/jigsaw"
    }],
    "require": {
        "tightenco/jigsaw": "dev-master"
    }
}
```

### DIFF
--- a/src/File/Filesystem.php
+++ b/src/File/Filesystem.php
@@ -8,8 +8,7 @@ class Filesystem extends BaseFilesystem
 {
     public function getFile($directory, $filename)
     {
-        $filePath = rtrim($directory, DIRECTORY_SEPARATOR).DIRECTORY_SEPARATOR.$filename;
-
+        $filePath = rtrim($directory, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $filename;
         return new SplFileInfo($filePath, $directory, $filename);
     }
 
--- a/src/File/Filesystem.php
+++ b/src/File/Filesystem.php
@@ -2,12 +2,15 @@
 
 use Illuminate\Filesystem\Filesystem as BaseFilesystem;
 use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\SplFileInfo;
 
 class Filesystem extends BaseFilesystem
 {
     public function getFile($directory, $filename)
     {
-        return iterator_to_array(Finder::create()->path(rtrim($directory, DIRECTORY_SEPARATOR).DIRECTORY_SEPARATOR.$filename)->append([]), false);
+        $filePath = rtrim($directory, DIRECTORY_SEPARATOR).DIRECTORY_SEPARATOR.$filename;
+
+        return new SplFileInfo($filePath, $directory, $filename);
     }
 
     public function putWithDirectories($file_path, $contents)
--- a/src/File/Filesystem.php
+++ b/src/File/Filesystem.php
@@ -7,7 +7,7 @@ class Filesystem extends BaseFilesystem
 {
     public function getFile($directory, $filename)
     {
-        return iterator_to_array(Finder::create()->files()->name($filename)->in($directory), false)[0];
+        return iterator_to_array(Finder::create()->path(rtrim($directory, DIRECTORY_SEPARATOR).DIRECTORY_SEPARATOR.$filename)->append([]), false);
     }
 
     public function putWithDirectories($file_path, $contents)
--- a/src/SiteBuilder.php
+++ b/src/SiteBuilder.php
@@ -20,8 +20,14 @@ class SiteBuilder
 
     public function build($source, $dest, $siteData)
     {
-        $this->prepareDirectories([$this->cachePath, $dest]);
-        $outputFiles = $this->writeFiles($source, $dest, $siteData);
+        $this->prepareDirectory($this->cachePath);
+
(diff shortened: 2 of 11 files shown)


damiani:
There are some great ideas here that would definitely be helpful for larger sites. It might even be worth building the complete site to a temp directory, then copying the temp directory to `build_production` upon completion, to (1) minimize downtime, and (2) protect `build_production` if something goes wrong during build. 

Can you explain the speed/memory difference? Is it solely the result of using `path()` rather than `files()` when calling the Symfony `Finder`?

abdumu:
Yes, my implementation will only clean build folder after the source generation ends which will prevent an empty production build folder.

The slowness is from Finder class, and my last commit I used ` Symfony\Component\Finder\SplFileInfo` directly for getting one fille and Finder is still good for `allFiles` method.  

I closed this commit because it didn't pass the tests, and now it does. (except for one test which is still impassable by original code)

damiani:
Excellent, thanks for updating. I'll take a look!

src/SiteBuilder.php
```diff
@@ -94,7 +145,7 @@ private function getMetaData($file, $baseUrl)
         $filename = $file->getFilenameWithoutExtension();
         $extension = $file->getFullExtension();
         $path = rightTrimPath($this->outputPathResolver->link($file->getRelativePath(), $filename, $file->getExtraBladeExtension() ?: 'html'));
-        $url = rightTrimPath($baseUrl) . '/' . trimPath($path);
+        $url = rightTrimPath($baseUrl).'/'.trimPath($path);
```
damiani: Keep spaces around `.`

src/SiteBuilder.php
```diff
@@ -55,15 +75,46 @@ private function cleanup()
...
+        $files = $files->map(function ($file) use ($destination, $progressBar) {
+            $progressBar->advance();
+
             return $this->writeFile($file, $destination);
         });
+
+        $progressBar->finish();
+        $this->consoleOutput->writeln('');
+
+        return $files;
```
damiani: In my testing, this second progress bar isn't all that useful, since writing happens very quickly. I would remove it.

src/File/Filesystem.php
```diff
@@ -2,12 +2,15 @@
...
 use Illuminate\Filesystem\Filesystem as BaseFilesystem;
 use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\SplFileInfo;
 
 class Filesystem extends BaseFilesystem
 {
     public function getFile($directory, $filename)
     {
-        return iterator_to_array(Finder::create()->files()->name($filename)->in($directory), false)[0];
+        $filePath = rtrim($directory, DIRECTORY_SEPARATOR).DIRECTORY_SEPARATOR.$filename;
```
damiani: Add spaces around `.`

cjf:
I'll just throw in my 2c here...

Just tested this on my site which is ~4k pages.  Vanilla jigsaw builds it in 1:12.  This PR builds it in 34 seconds.  Diff showed the site building identically.

I'd just be incredibly pleased to see this integrated.

damiani:
Great...we're working on some other optimizations that will reduce that oven further.

abdumu:
I'm gonna close this PR.
Jigsaw is not meant for large number of files (crashes due to memory consumption) , and my upcoming changes won't be welcomed because I'm going and I did some change to the 'core' codes.

For those who face performance/speed issues with current Jigsaw, you are more than welcome to use/fix my forked repo.

Thanks.

cjf:
Is this actually dead?  I thought progress was being made?

damiani:
Nope, it's not dead. We'll be incorporating the `SpfFileInfo` change that @abdumu suggested here, along with a few other performance improvements we've been working on, in the next Jigsaw release.

abdumu:
@cjf  

I don't think it's dead, I closed the repo because it's now a different code that what described earlier and now it has its own package for those who wants a version that focus on performance and speed: 

https://github.com/abdumu/jigsaw

enjoy.

mattstauffer:
@abdumu Is there a performance improvement you think needs to be made that has not been either made or planned? This feels like a bit of a weird interaction.

abdumu:
@mattstauffer what "weird interaction" ?

mattstauffer:
@abdumu Keith has made it clear he's intending to including changes targeted at large numbers of files, and that he's incorporating your suggestion. You then closed the PR saying that Jigsaw won't be making the changes we said we're making, and you're now maintaining a fork because you're saying Jigsaw isn't focusing on the things that Keith just said we're going to address.

My assumption is that we're just having a miscommunication here, or everyone else reading this understands something I don't, which is why I'm asking you for clarification. 

Thanks!

abdumu:
@mattstauffer I said "I closed [it]" because "it's now a different code than what described earlier" and " [it] changes ... the 'core' codes" ..

mattstauffer:
@abdumu OK, cool. :) Thanks for clarifying that it was about a difference in opinion of *how* to address an issue, not *whether*. <3

PR #235, opened 2018-07-27
Drop non-generated CSS file

mattstauffer:
Since we have Mix out of the box, we can drop the old default `main.css` file, as it's just confusing at this point.


mattstauffer:
@dborsatto Yep, it does... sometimes it's just hard to notice that in the UI.

image

dborsatto:
Oh, you're right! Sorry 🙂

PR #237, opened 2018-07-27
make Jigsaw main class macroable

drweissbrot:
This PR changes the main Jigsaw class to be macroable, using the default Laravel trait, as provided by illuminate/support.

I personally would like to use this to access the protected properties of the Jigsaw object. For instance, this could be useful to modify (or add) Collections.

Using macros appears to be the only (or at least, best) option to accomplish this, since the Jigsaw architechture doesn't really allow modifying or appending to the Jigsaw class anywhere else.

### DIFF
--- a/src/Jigsaw.php
+++ b/src/Jigsaw.php
@@ -1,9 +1,12 @@
 <?php namespace TightenCo\Jigsaw;
 
 use TightenCo\Jigsaw\File\Filesystem;
+use Illuminate\Support\Traits\Macroable;
 
 class Jigsaw
 {
+    use Macroable;
+
     public $app;
     protected $env;
     protected $outputPaths;


damiani:
This should possible using an [Event Listenter](http://jigsaw.tighten.co/docs/event-listeners/) in `bootstrap.php`, and listening to either the `beforeBuild` or `afterCollections` events, depending on what you're trying to accomplish. Make sure to update Jigsaw to the latest release (v1.2.7), which includes some fixes for collection handling in event listeners.

drweissbrot:
I've probably not been as clear as I should've been. I'm trying to add a separate collection of items that are already in another collection, i. e. I want to categorize my blog posts into a separate collection (e. g. items from Collection posts with the attribute category = "coding" into Collections posts_coding).

Trying to accomplish this only using the event system, I came up with this:
```php
$events->afterCollections(function (Jigsaw $jigsaw) {
    $posts = $jigsaw->getCollection('posts');
    $categories = $articles->mapToGroups(function ($item, $key) {
        return [$item['category'] => $item];
    });

    foreach ($categories as $category => $items) {
        $jigsaw->setConfig('collections.posts_' . str_slug($category, '_'), compact('items'));
    }
});
```
This doesn't work though, since Jigsaw is already done creating collections at the point this event is fired. I have to wait for this event though and can't just use another one, since I need the normal posts Collection already done.

Using macros, this appears to work just fine:
```php
Jigsaw::macro('getSiteData', function () {
    return $this->siteData;
});

Jigsaw::macro('setSiteData', function (SiteData $siteData) {
    $this->siteData = $siteData;
});

$events->afterCollections(function (Jigsaw $jigsaw) {
    $siteData = $jigsaw->getSiteData();

    $categories = $siteData->get('posts')->mapToGroups(function ($item, $key) {
        return [$item['category'] => $item];
    });

    foreach ($categories as $category => $items) {
        $siteData->put('posts_' . str_slug($category, '_'), $items);
    }

    $jigsaw->setSiteData($siteData);
});
```

damiani:
Can you give me a little more detail about how you're planning on using the categorized collections? Do you need to duplicate the actual blog posts, or are you just trying to build categorized "index" pages, like on [this demo page](https://tightenco.github.io/jigsaw-collections-demo/categories/news/)? If so, there might be a slightly easier way to solve this than by modifying `siteData`.

drweissbrot:
Yes, what I'm trying to do is essentially the same as in the example. I want all posts in a single collection, but I also want some pages only showing posts that are in a specific category. That would of course work like in the demo, but I'd also like to have these category pages paginated. And as far as I know, there is no way to tell Jigsaw to filter a collection before paginating in the front matter, or am I wrong on that?

damiani:
The `setConfig()` method accomplishes the same thing (it was just updated the other day, in v1.2.6), so you can use that without needing to directly modify `siteData`. I just pushed a small fix (v1.2.8) that fixes a couple issues that `setConfig()` would have when trying to do what you're doing...so if you upgrade, and use this code, you should be good to go:

```php
$events->afterCollections(function (Jigsaw $jigsaw) {
    $categories = $jigsaw->getCollection('posts')->mapToGroups(function ($item) {
        return [$item['category'] => $item];
    })->each(function ($items, $category) use ($jigsaw) {
        $jigsaw->setConfig('posts_' . str_slug($category, '_'), $items);
    });
});
```

(I'm assuming you have a `posts_coding.blade.php` layout that grabs the new `posts_coding` collection and paginates it.)

Can you upgrade Jigsaw to v1.2.8, and then test this out?

drweissbrot:
Tried it out with 1.2.8, and it appears to be working without any problems. Thanks for your time!

Though I still think it might be worth considering to have the class macroable, I'm assuming you don't want that and will close this PR.

damiani:
Excellent! The reason I don't want to make it macroable, or change those 5 protected properties to public, is that there's enough magic going on in some of those (`siteData` in particular) that people who attempt to modify them manually are likely to run into dragons.

PR #242, opened 2018-08-04
adding php-cs-fixer to development

linuxjuggler:
Using PHP CS Fixer we can provide some guide lines for how the code should be formatted, this way running a simple command `composer run fix` would make it easier to fix.

The rules which I've set within `.php_cs` file are just a sample I like to always have, and for sure we can have more or edit them to make them more better/suited for the project.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,22 @@
+<?php
+ $finder = PhpCsFixer\Finder::create()
+    ->name('*.php')
+    ->in(__DIR__.DIRECTORY_SEPARATOR.'src')
+    ->in(__DIR__.DIRECTORY_SEPARATOR.'tests');
+ return PhpCsFixer\Config::create()
+    ->setRiskyAllowed(true)
+    ->setRules([
+        '@PSR2' => true,
+        '@Symfony' => true,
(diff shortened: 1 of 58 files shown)


damiani:
Fantastic!

Here's the rules I'd prefer to use, alphabetized for good measure :) 

```php
'@PSR2' => true,
'@Symfony' => true,
'align_multiline_comment' => ['comment_type' => 'all_multiline'],
'array_syntax' => ['syntax' => 'short'],
'concat_space' => ['spacing' => 'one'],
'method_argument_space' => ['on_multiline' => 'ensure_fully_multiline'],
'not_operator_with_successor_space' => true,
'phpdoc_annotation_without_dot' => true,
'phpdoc_order' => true,
'trailing_comma_in_multiline_array' => true,
'yoda_style' => false,
```

When updating these rules, you'll need to revert the changes that php-cs-fixer made using your previous config, and re-run it again; some of the changes it made won't be "un-done" otherwise.

linuxjuggler:
Thanks, 

I think I'll have to close this one and send a request for the new one.

thanks @damiani

PR #243, opened 2018-08-05
adding PHP CS Fixer

linuxjuggler:
Using PHP CS Fixer we can provide some guidelines to show how the code should be formatted, this way running a simple command `composer run fix` would make it easier to fix.

The rules which I've set within the `.php_cs` file is based on the following comment https://github.com/tightenco/jigsaw/pull/242#issuecomment-410484789


damiani:
Thanks for this contribution @linuxjuggler !

PR #244, opened 2018-08-06
remove composer.lock from gitignore

linuxjuggler:
based on getcomposer.org recommendation at https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control it is important to have `composer.lock` within git.

> Committing this file to VC is important because it will cause anyone who sets up the project to use the exact same versions of the dependencies that you are using. Your CI server, production machines, other developers in your team, everything and everyone runs on the same dependencies, which mitigates the potential for bugs affecting only some parts of the deployments. Even if you develop alone, in six months when reinstalling the project you can feel confident the dependencies installed are still working even if your dependencies released many new versions since then. (See note below about using the update command.)

I know that there are some debate whether we should have this within frameworks or not, but if am not mistaken `Jigsaw` is an application not a framework.

More info about why we should have `composer.lock` within git was mentioned here https://www.engineyard.com/blog/composer-its-all-about-the-lock-file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 /vendor
 .DS_Store
-composer.lock
 _tmp/
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,3479 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
+        "This file is @generated automatically"
+    ],
+    "content-hash": "6f8e2954d1d627dc2dd1c4aad213a62a",
+    "packages": [
+        {
+            "name": "doctrine/inflector",
+            "version": "v1.3.0",
+            "source": {
+                "type": "git",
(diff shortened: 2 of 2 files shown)


linuxjuggler:
We will always have a debate weather it is an application or a framework, as for example I cant say that `Laravel Mix` is a framework, its a tool, the same thing with `Jigsaw` in my opinion,  it is just a set of tool that any user can use it, remember each one of us can define a different specification for the word `end-users`.

What can help us is to define `who is the end-user for Jigsaw`, a designer, a programmer or anyone with few knowledge of markup?

pedzed:
> What can help us is to define who is the end-user for Jigsaw, a designer, a programmer or anyone with few knowledge of markup?

The end-user is the user at the end? I.e. the one who visits and uses the website/web application. The programmer or designer just writes it for the end-user.

linuxjuggler:
Based on that the end user has no interest in jigsaw or what files we
include or exclude from it.and they only care about the website they visit
with the mabojambo style and layout

From my mobile, Please excuse any awkward autocorrections or typos.

Issue #251, opened 2018-08-21
Proposal: $page->excerpt() on CollectionItem


raniesantos:
**If you guys (Matt, Keith, other maintainers) think this is a good idea, I will write it myself and make a PR.**

**I have this helper in my config, but I think it is a good SSG feature.**

Here's a rough example of what it would look like:

In `CollectionItem.php`:
```php
<?php

use Illuminate\Support\Str;

public function excerpt($limit = 250, $end = '...')
{
    $newlines = ["\r", "\n", "\r\n"];
    return Str::limit(str_replace($newlines, ' ', strip_tags($this->_content)), $limit, $end);
}
```

If one of the below actions seems unnecessary I can just exclude it from the code.

- strips tags
- removes line breaks so that it outputs content as a straight line in blade/html (this is specifically useful for outputting part of the content in meta tags)
- limits characters, configurable (I looked up average meta description lengths and decided 250 is a good default)
- appends an ellipsis, also configurable

**Example of real-world use (SEO and social sharing):**

```handlebars
<!-- code is specific to posts/markdown pages -->
<meta name="description" content="{{ $page->excerpt() }}">
<meta property="og:description" content="{{ $page->excerpt() }}">
<meta name="twitter:description" content="{{ $page->excerpt() }}">
```

It can also be useful in a blog index where posts are looped and content excerpts are shown.

hellocosmin:
Wouldn't this be the case for a [helper method](http://jigsaw.tighten.co/docs/helper-methods/)?

raniesantos:
I do have this as a helper method in my config. But I think that this is a feature that fits well in a SSG.

SSGs are slowly becoming more relevant and less of just hobby projects. So more people would want to add stuff like this into their sites.

```html
<meta name="description" content="{{ $page->excerpt() }}">
<meta property="og:description" content="{{ $page->excerpt() }}">
<meta name="twitter:description" content="{{ $page->excerpt() }}">
```

Having this built in helper would be helpful.

damiani:
I think this would be better suited as a helper method in `config.php`, where it can be more easily user-configurable (preserving certain tags, for instance). 

That said, it might be worthwhile adding this to the default `config.php` stub as a sample helper method, since it's a pretty good demonstration of how helper methods work.

If we do that, I would make the function respect word boundaries, either using `preg_replace('/\s+?(\S+)?$/' ... )` or `strtok(wordwrap($str, length, "...\n"), "\n")`

[...]

PR #255, opened 2018-08-23
Added `isMarkdown()` method to `PageVariable` class

raniesantos:
This PR adds an `isMarkdown()` method in the `PageVariable` class.

___

The number one purpose of this method is to be used with `getContent`.

The docs state that `getContent()` *is available for Markdown files only*.

This method will help in situations where you need to determine if `getContent()` is usable on certain pages. Specifically when the relevant code is in a layout that may be used by both blade and markdown files (like the `master.blade.php` layout).

___

Besides the specific purpose mentioned above, this method can have many uses considering that a SSG will likely contain tons of markdown files.

___

My usage:

```html
<meta name="description" content="{!! $page->isMarkdown() ? $page->excerpt() : $page->description !!}">
<meta property="og:description" content="{!! $page->isMarkdown() ? $page->excerpt() : $page->description !!}">
<meta property="og:type" content="{{ $page->isMarkdown() ? 'article' : 'website' }}">
<meta name="twitter:description" content="{!! $page->isMarkdown() ? $page->excerpt() : $page->description !!}">
```

Please also see [issue 251](https://github.com/tightenco/jigsaw/issues/251) which is related to this snippet

### DIFF
--- a/src/PageVariable.php
+++ b/src/PageVariable.php
@@ -54,6 +54,11 @@ class PageVariable extends IterableObject
         return $this->_meta->url;
     }
 
+    public function isMarkdown()
+    {
+        return in_array($this->getExtension(), ['markdown', 'md', 'mdown']);
+    }
+
     protected function missingHelperError($functionName)
     {
         return 'No function named "' . $functionName . '" was found in the file "config.php".';


damiani:
I think this would be better suited as a helper method in `config.php`; plus, the `getContent()` restriction on Blade files is going away soon.

Issue #191, opened 2018-03-08
Jigsaw Scalability Max # of Markdown Files


dgursh:
Hello all, I am wondering what the scalability of jigsaw is? What is the most number of markdown files / pages has the build command been tested with? I'm currently working on a project with over 1K+ markdown files and I'm running into issues when trying to build.

```
Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 32768 bytes)
```

Not sure if it has to do with buggy foreach loops that I may have written by mistake @damiani @adamwathan can you chime in here?

Thanks.

dgursh:
I seemed to have fixed the bug, by setting the memory limit in the php.ini file to the following:

```
memory_limit = 256M
```

I know this isn't the best solution. But, I checked my code and it wasn't buggy. So, I guess due to the number of files I was working with I needed more memory.

Hope this helps someone in the future.

damiani:
We're working on testing the limits to see where we might be able to optimize things for larger sites. Do you have a repo you could share with us that we can test against?

dgursh:
@damiani I do not have a test repo currently available, I will see if I can make one.

dgursh:
@damiani Returning to this thread. I am running into issues with builds of 6,000+ pages. It takes around 5 minutes each time I build the site. 

My follow up question to you: 

Is there a way to only build the files that change. For example, if I make a small change in one markdown file, is there a way I can only build up that one new file?

damiani:
6,000+ is a lot of pages :)

There's no way to do this at the moment, and it's a tricky feature to add because a change to a single markdown file could potentially have repercussions on every other page (an added collection item that needs to appear in a menu, for instance, or a changed variable in a YAML header).

(Is your repo private? I'm looking into #226 and would love to test against a real-world example of a very large Jigsaw site.)

dgursh:
@damiani Unfortunately this is a private repo, but what I can do is spin up a test website that has this many pages to test on.

#226 is interesting and it looks like it speeds up the site build.

I want to stick with Jigsaw because I am using blade in this project and the ease of portability between Jigsaw and Laravel is a must. However, I've been looking into Hugo (Go) because it takes < 1ms to build a file instead of 1s with Jigsaw. This performance boost is necessary.

damiani:
[...]

PR #260, opened 2018-08-28
Performance improvements

damiani:
(Addresses https://github.com/tightenco/jigsaw/issues/191)

This PR implements a variety of optimizations to improve the performance of running `jigsaw build`. These optimizations are of greatest benefit to large Jigsaw sites (> 1,000 pages or so).

Big thanks to @abdumu who suggested a couple of these improvements in https://github.com/tightenco/jigsaw/pull/226:
- replacing use of Symfony Finder w/ `SplFileInfo`
- using a fake when building the view `Factory`, rather than a mock created by Mockery

In addition, `MarkdownHandler` has been optimized to reduce the amount of I/O involved:
- the number of temporary files created for each Markdown page has been reduced from 4 to 1
- temporary filenames are hashed to take advantage of Blade's built-in caching

## Caching

Finally, support for caching has been added. Caching can be turned on by adding the `--cache` or `-c` option to the `jigsaw build` command, or by adding `'cache' => true` to your `config.php` file. Using the cache will add a `cache` directory to your project root, and will attempt to use those files to speed up subsequent builds. 

> Note that the **first** run with `-c` will not _use_ the cache when building the site, but will _create_ the cache files used in subsequent builds. 

> Running `jigsaw build` without the `-c` option will delete the `cache` directory.

## Progress indicators

The console output has also been improved, to include a progress indicators for the collection processing and site building steps. To view the progress bars, add the `-v` option to `jigsaw build`. To silence output entirely, pass the `-q` (or `--quiet`) option.

jigsaw-cache-output

## Benchmarks

These improvements are most noticeable on large Jigsaw sites that use Markdown heavily:

| 6,000 files           |             |
|-----------------------|-------------|
| Original              | 1.6 minutes |
| Optimized             | 11 seconds  |
| Optimized, with cache | 7 seconds   |

| 1,000 files           |            |
|-----------------------|------------|
| Original              | 14 seconds |
| Optimized             | 1 second   |
| Optimized, with cache | .6 seconds |

The changes in this PR substantially reduce memory consumption during `build`. For the 6,000-page demo site, memory consumption was reduced from 180 MB (!) to 39 MB; for the 1,000-page site, memory consumption was reduced from 17 MB to 7 MB.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /vendor
 .DS_Store
 composer.lock
+.phpunit.result.cache
 _tmp/
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
+/cache
 /vendor
 .DS_Store
 composer.lock
 .phpunit.result.cache
-_tmp/
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,6 @@
         "illuminate/support": "^5.6",
         "illuminate/view": "^5.6",
         "symfony/console": "~3.0|^4.0",
-        "mockery/mockery": "^1.0.0",
         "mnapoli/front-yaml": "^1.5",
         "symfony/yaml": "^4.0",
         "erusev/parsedown-extra": "^0.7.1"
@@ -36,12 +35,12 @@
         "psr-4": {
             "Tests\\": "tests/"
         }
-    },        
+    },
     "bin": [
         "jigsaw"
     ],
     "scripts": {
-        "fix" : "php vendor/bin/php-cs-fixer fix --using-cache=no"   
+        "fix" : "php vendor/bin/php-cs-fixer fix --using-cache=no"
     },
     "prefer-stable": true,
     "config": {
--- a/jigsaw-core.php
+++ b/jigsaw-core.php
@@ -12,10 +12,11 @@ use Mni\FrontYAML\Bridge\Symfony\SymfonyYAMLParser;
 use Mni\FrontYAML\Markdown\MarkdownParser;
 use Mni\FrontYAML\Parser;
 use Mni\FrontYAML\YAML\YAMLParser;
-use TightenCo\Jigsaw\Collection\CollectionPaginator;
 use TightenCo\Jigsaw\CollectionItemHandlers\BladeCollectionItemHandler;
 use TightenCo\Jigsaw\CollectionItemHandlers\MarkdownCollectionItemHandler;
+use TightenCo\Jigsaw\Collection\CollectionPaginator;
 use TightenCo\Jigsaw\Events\EventBus;
+use TightenCo\Jigsaw\Events\FakeDispatcher;
 use TightenCo\Jigsaw\File\BladeDirectivesFile;
 use TightenCo\Jigsaw\File\ConfigFile;
 use TightenCo\Jigsaw\File\Filesystem;
@@ -105,7 +106,7 @@ $container->bind(Factory::class, function ($c) use ($cachePath) {
 
     $finder = new FileViewFinder(new Filesystem, [$cachePath, $c['buildPath']['source']]);
 
-    return new Factory($resolver, $finder, Mockery::mock(Dispatcher::class)->shouldIgnoreMissing());
+    return new Factory($resolver, $finder, new FakeDispatcher());
 });
 
 $container->bind(ViewRenderer::class, function ($c) {
--- a/jigsaw-core.php
+++ b/jigsaw-core.php
@@ -15,6 +15,7 @@ use Mni\FrontYAML\YAML\YAMLParser;
 use TightenCo\Jigsaw\CollectionItemHandlers\BladeCollectionItemHandler;
 use TightenCo\Jigsaw\CollectionItemHandlers\MarkdownCollectionItemHandler;
 use TightenCo\Jigsaw\Collection\CollectionPaginator;
+use TightenCo\Jigsaw\Console\ConsoleOutput;
 use TightenCo\Jigsaw\Events\EventBus;
 use TightenCo\Jigsaw\Events\FakeDispatcher;
 use TightenCo\Jigsaw\File\BladeDirectivesFile;
@@ -62,6 +63,10 @@ $container->instance('buildPath', [
 
 $container->bind('config', function ($c) {
     return (new ConfigFile($c['cwd'] . '/config.php'))->config;
+});
+
+$container->singleton('consoleOutput', function ($c) {
+    return new ConsoleOutput();
 });
 
 $container->bind('outputPathResolver', function ($c) {
@@ -114,12 +119,12 @@ $container->bind(ViewRenderer::class, function ($c) use ($bladeCompiler) {
     return new ViewRenderer($c[Factory::class], $bladeCompiler);
 });
 
-$container->bind(BladeHandler::class, function ($c) {
-    return new BladeHandler($c[TemporaryFilesystem::class], $c[FrontMatterParser::class], $c[ViewRenderer::class]);
-});
-
 $container->bind(TemporaryFilesystem::class, function ($c) use ($cachePath) {
     return new TemporaryFilesystem($cachePath);
+});
+
+$container->bind(BladeHandler::class, function ($c) {
+    return new BladeHandler($c[TemporaryFilesystem::class], $c[FrontMatterParser::class], $c[ViewRenderer::class]);
 });
 
 $container->bind(MarkdownHandler::class, function ($c) {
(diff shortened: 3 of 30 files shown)


cjf:
Seems to be passing my initial tests with flying colors.  Build time has dropped from 1:07 to < 30s!  

The cache option doesn't seem to improve anything, however.  The cache folder is being created and populated, but nothing seems to change.

First build with -c...

` » time ../jigsaw2/jigsaw build -c -v

Build time: 25.01 seconds (using cache)
Site build successfully!

../jigsaw2/jigsaw build -c -v  20.76s user 4.00s system 98% cpu 25.144 total
`

Follow up build changing nothing...

`» time ../jigsaw2/jigsaw build -v -c

Build time: 25.18 seconds (cache was created)
Site build successfully!

../jigsaw2/jigsaw build -v -c  21.04s user 3.83s system 98% cpu 25.321 total
`

damiani:
@cjf The 2nd output says "cache was created", which means that it wasn't used for that build, but would be for the next. Can you run it a couple times with the `-c` option?

Also:
1. Does your site use primarily Markdown or Blade files for collection items?
2. Would you be willing to share a copy of your site with me privately for testing?

abdumu:
I must say, this is well done .. I tested it and bravo!


(4000 files)

with '-c'
```
Build time: 7.99 seconds (cache was created)
Site build successfully!

0:08.08 real,	5.07 user,	2.88 sys,	0 amem,	139568 mmem
```

then again with '-c'
```
Build time: 5.02 seconds (using cache)
Site build successfully!

0:05.09 real,	2.94 user,	1.96 sys,	0 amem,	137664 mmem
```

finally, without '-c'
```
Build time: 9.05 seconds
Site build successfully!

0:09.13 real,	5.18 user,	3.76 sys,	0 amem,	139352 mmem
```

This is an example of how a perspective of a person could change in 28 days ! 😉
https://github.com/tightenco/jigsaw/issues/191#issuecomment-409259652


Good job. will be waiting the next release.

cjf:
@damiani

Re: output of those runs, I must have copied them in backwards.  I've been testing again today, with the same results.

> Does your site use primarily Markdown or Blade files for collection items?

Primarily .blade.md.

> Would you be willing to share a copy of your site with me privately for testing?

I wish I could share my site, but that's not possible at the moment.  It's not mine to share.  I'll see what I can do, though.

cjf:
I've also noticed a bug in how the markdown is processed.  This new version pukes on mailto links that used to work.

This now fails...

`[test@test.com](mailto:test@test.com)`

`In 7b7733b9c50b05c5bc197815bcc1b5af745906de.php line 1:

  syntax error, unexpected '&'

`

The code it produces is...

`<?php $__env->startSection('content'); ?><p><a href="mailto:test<?php echo e(&#039;@&#039;); ?>test.com">test<?php echo e('@'); ?>test.com</a></p><?php $__env->stopSection(); ?>
<?php echo $__env->make('_layouts.master', array_except(get_defined_vars(), array('__data', '__path')))->render(); ?>`

damiani:
@cjf Ah yeah, `.blade.md` collection items don't take advantage of caching the way straight `.md` files do. I'll see if I can make some improvements there as well.

And good to know about the bug. Is that in a `.md` or a `.blade.md` file?

cjf:
Plain markdown.  It works for blade markdown

dgursh:
@damiani 

> With The SideBar that has Foreach Loops on large collections
> Build time: 39.98 seconds (cache was created)
> Build time: 35.85 seconds (using cache)
> Build time: 33.83 seconds (using cache)
> Build time: 37.23 seconds (using cache)

> Without Sidebar 
> Build time: 15.44 seconds (cache was created)
> Build time: 13.78 seconds (using cache)
> Build time: 12.57 seconds (using cache)
> Build time: 14.18 seconds (using cache)

As you can see significant improvements when I remove the logic around the collections. If you'd like I can add something similar to the test repo.

dgursh:
@damiani I added a sidebar with a foreach loop using the lesssons collection (> 6000) to show you what I mean. It increases the build time to be minutes. Removing that single foreach loop will decrease the build time to ~11 seconds. I think it's important we provide a fix for this.

https://github.com/dgursh/jigsaw-bugs

raniesantos:
> Finally, support for caching has been added. Caching can be turned on by adding the --cache or -c option to the jigsaw build command, or by adding 'cache' => true to your config.php file

Shouldn't *anything that makes the build faster* be ON by default?

dgursh:
@damiani I want to point one more special use case with my application's sidebar that I think we should consider.

```
@foreach ($lessons->where('course', $page->course) as $lesson)
	            <li><a href="{{ $lesson->getUrl().'/' }}" class="{{ $page->active($term->getUrl()) }}"> {{ $term->title }}</a></li>
	        @endforeach
```
I am returning some special class styling to the sidebar item that the current page is on with the active property defined in my config file.

So, the majority of this foreach loop is the same, with one exception the class on one item that is being applied. Still, I don't think this should be a case that requires the entire collection to be run.

Any ideas on how we would we solve this special case? I'm happy to put an example in the repo.

damiani:
@dgursh Nope just haven't had a chance to look into the sidebar yet; I'll be looking at it later this week, though I plan on addressing @cjf 's bug first (which is related to `@` escaping). 

You are, of course, more than welcome to dive into the code in the meantime as you wish!

dgursh:
> (Moved from #191)
> 
> from @dgursh:
> 
> > What I'm noticing is that my website runs foreach and eloquent queries using the collections (of > 1000 items) and embeds them on every single page as in the case of a sidebar navigation.
> 
> > Can you think of any way where we can speed up this use-case? It shouldn't have to perform these massive foreach queries on big collections when building up every page. Maybe a possible solution is a pre-build step for any foreach clause that uses a collection.
> 
> I can think of a couple ways to deal with:
> 
> * build the sidebar nav content in an `afterCollections` [event listener](http://jigsaw.tighten.co/docs/event-listeners/), write the nav HTML to a config variable, and `{!! !!}` that on your template
> * I've toyed with creating a `@once` Blade directive that you could wrap around expensive `@foreach` clauses like this one, that would tell Jigsaw to pre-render content that doesn't change from page to page. (Related to that, now that we have the concept of caching, we can also add something like `@cache(7)` to cache content for a particular amount of time.)
> 
> Your repo is private, so I can't test any of these approaches on your codebase. But could you benchmark the build time (with caching on) with the side nav removed, to see just how much of your build time is occupied by that `@foreach` statement?

@damiani I've been digging into the code and trying to implement the event listener. Do you have your code for your "@once" blade directive I can take a look at?

Thanks!

dgursh:
@damiani how do you access the $page variable in bootstrap.php afterCollections?

damiani:
You would target a collection via `$jigsaw->getCollection('{collection_name}')`, then iterate over that. Each item is then accessible in the same way `$page` is in your templates.

For example:

```php
$events->afterCollections(function (Jigsaw $jigsaw) {
    $jigsaw->getCollection('posts')->each(function ($item) {
        var_dump($item->getUrl());
    });
});
```

dgursh:
@damiani I am trying to translate blade code into the afterCollections event listener like you suggested:

```
@foreach ($courses->where('topic', $page->topic) as $course)
    <li><a href="{{ $course->getUrl().'/' }}" class="{{ $page->active($course->getUrl()) }}">{{ ucwords($course->title) }}</a></li>
	@foreach ($lessons->where('course', $course->title) as $lesson)
    	<li><a href="{{ $lesson->getUrl().'/' }}" class="{{ $page->active($lesson->getUrl()) }}">{{ ucwords($lesson->title) }}</a></li>
    @endforeach
@endforeach
```
I need to access the page's data and not just the collections data for my where clauses. How would I do this?

damiani:
@dgursh There is no "current page that's being built" when you're in that event listener; you have access to the config, and the collection items, but the files themselves are processed in between `afterCollections` and `afterBuild`, and there's no event that gets fired once for every file. When I suggested `afterCollections` as a possible approach, I hadn't realized you were modifying that sidebar for each page; `$page->active($course->getUrl())` isn't possible inside an event listener.

dgursh:
@damiani Right, so I'm in a bit of bind then. I guess you cannot build the sidebar once, if it's technically different for every page based on the highlighting of the url. Any ideas on how I can get the highlighting in this case? Maybe javascript instead?

damiani:
Yeah, on a site like yours, JS is probably the best solution for styling the active page.

dgursh:
active page is easy. But how would I only show a subset of the lessons based on the page?

abdumu:
back for reporting.. 
I tried it today on 5000+ files and it crashed again.

PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 294912 bytes) in /Users/.../www/jigsaw/vendor/illuminate/support/Collection.php on line 1785

yet abdumu/jigsaw is still working.

I think it still needs some investigations.

keep up the good work.

dgursh:
> @damiani I added a sidebar with a foreach loop using the lessons collection (> 6000) to show you what I mean. Test it if you can to see what I mean...
> 
> It increases the build time to be minutes. Removing that single foreach loop will decrease the build time to ~11 seconds. I think it's important we provide a fix for this like the one you mentioned above especially since it is this slow for a simple for loop.
> 
> https://github.com/dgursh/jigsaw-bugs

@damiani was this ever solved? Foreach loops with very large collections on every page?

damiani:
@dgursh No, I haven't come up with a viable solution to optimize the handling of those `foreach` loops, at least not one that wouldn't require a fundamental change to the use of the Blade compiler in Jigsaw.

PR #264, opened 2018-08-30
Allow merging optional helpers file to config file

### LINKED ISSUE #263: Allow optional helpers.php file
### Proposal

I suggest enabling Jigsaw to check if a `helpers.php` file exists in the project root.

If true, merge the arrays in `config.php` and `helpers.php`.

If false, nothing breaks (similar to how the `blade.php` file is optional).

### Reason

Jigsaw's flexibility allows us to easily add any helper in the config file.

This is great, but generally a config file should consist of constant/static values. It would make sense to put helper functions in a separate file.

Users should have the option to use a separate file, but still be free to put everything in a single file if they prefer it that way.

raniesantos:
Closes #263

Copied from issue:

> ### Proposal
> 
> I suggest enabling Jigsaw to check if a `helpers.php` file exists in the project root.
> 
> If true, merge the arrays in `config.php` and `helpers.php`.
> 
> If false, nothing breaks (similar to how the `blade.php` file is optional).
> 
> ### Reason
> 
> Jigsaw's flexibility allows us to easily add any helper in the config file.
> 
> This is great, but generally a config file should consist of constant/static values. It would make sense to put helper functions in a separate file.
> 
> Users should have the option to use a separate file, but still be free to put everything in a single file if they prefer it that way.

Additional note:

I decided to add the code in `ConfigFile.php` because *although helpers is a separate file, it is essentially just an extension of the config*.

### DIFF
--- a/jigsaw-core.php
+++ b/jigsaw-core.php
@@ -60,7 +60,7 @@ $container->instance('buildPath', [
 ]);
 
 $container->bind('config', function ($c) {
-    return (new ConfigFile($c['cwd'] . '/config.php'))->config;
+    return (new ConfigFile($c['cwd'] . '/config.php', $c['cwd'] . '/helpers.php'))->config;
 });
 
 $container->bind('outputPathResolver', function ($c) {
--- a/src/File/ConfigFile.php
+++ b/src/File/ConfigFile.php
@@ -6,9 +6,12 @@ class ConfigFile
 {
     public $config;
 
-    public function __construct($file_path)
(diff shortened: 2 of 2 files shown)


Log1x:
Would definitely like to see this simple change merged.

damiani:
I feel like this can be done more explicitly by just including additional files in your `config.php` and merging them:

```
return [
    // ... your config array
] + include('helpers.php');
```

This way, you can organize additional config or helper files however you'd like.

PR #267, opened 2018-09-05
Adds support to inline asset (css/js)

erickpatrick:
depending on the size of your CSS or JS, you may want to inline them
into the style or script tag of the head of your pages so you save
one request and avoid blocking the rest of the page from loading.

Due to that, we added the `inline` function that will be used in
conjunction to the `mix` function to properly get the content
of your asset file and print it where we want it.


erickpatrick:
By the way, usage is

`{{ inline(mix('css/main.css', 'assets/build')) }}`

erickpatrick:
Updated to latest version of jigsaw

src/Support/helpers.php
```diff
@@ -116,3 +116,16 @@ function dd(...$args)
         die(1);
     }
 }
+
+function inline($assetPath)
+{
+    preg_match('/^\/assets\/build\/(css|js)\/main\.(css|js)/', $assetPath, $matches);
```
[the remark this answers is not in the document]
erickpatrick: This can be changed to a ‘*’ instead of ‘main’ and it should work just fine. I can test this change and update the PR

m1guelpf:
```suggestion
    preg_match('/^\/assets\/build\/(css|js)\/.*\.(css|js)/', $assetPath, $matches);
```
This is what worked for me locally

erickpatrick: Thanks @m1guelpf

molerat619:
Can you support me with https://github.com/tightenco/jigsaw/issues/471 please?

PR #277, opened 2018-10-02
Replace special characters with ASCII replacements. Fixes #269.

### LINKED ISSUE #269: Jigsaw slugify removes accented characters
Using the dash slugify of Jigsaw removes accented characters.

Example:
```
'posts' => [                                                             
            'extends' => '_layouts.post',                                        
            'path' => '{language}/{collection}/{-title}',
             'items' => [
                  'title' => 'Fútbol solidario'
]
```

The generated slugified title is `.../es/posts/ftbol-solidario/` with the `ú` dropped.

Would it be possible to instead of dropping the accented character just remove the accent, this would make it easier to use the Laravel `str_slug` function which removes accents and not the whole character.

bhulsman:
I'm not sure if I should add tests, would love to though! I just couldn't find tests that test similar functionality. Existing tests all seem to run fine.

Basically I simply ported the Laravel 5.7 Str::slug() functionality which supports replacing special characters to ASCII ones. I incorporated the changes for allowing dots.

Feedback/suggestions/anything is welcome!


damiani:
@bhulsman This is great! I added some tests, and in the process, realized that in some cases, the `filterInvalidCharacters` method was filtering out international characters before they could be transliterated. So I removed that (which isn't necessary anyway, since it's covered by the second `preg_replace` in `slug()` anyway), and moved the `ascii()` call to `cleanOutputPath()` so that it covers all circumstances.

Take a look and let me know if you have any thoughts, then I'll merge this in!

bhulsman:
Ah nice! I did not realize the `slug` method was already filtering those characters. Tests cover the special character cases from what I can think of, looking good!

damiani:
@bhulsman Thanks for your contribution!

PR #279, opened 2018-10-05
Add confirmation for custom destination paths

### LINKED ISSUE #276: Present user with warning when building to non-standard destination
(For context, see https://github.com/tightenco/jigsaw/issues/222)

Add a console warning when running `jigsaw build` if the default `destination` path has been modified (from the default `build_*` in the project directory. 

Because the destination directory is deleted as part of the build process, at the moment there's the potential for users to unwittingly delete a directory inside their project, or their entire project, if they get the destination path wrong. 

We would also need to provide the ability to override or silence this warning, so that `jigsaw build` will function correctly when watching files with Webpack.

mattisbones:
This commit resolves #276 by presenting the user with a confirmation prompt when the build destination is set to a custom path. A `--no-warnings` option can be passed to the command to bypass this check.

### DIFF
--- a/src/Console/BuildCommand.php
+++ b/src/Console/BuildCommand.php
@@ -23,7 +23,8 @@ class BuildCommand extends Command
         $this->setName('build')
             ->setDescription('Build your site.')
             ->addArgument('env', InputArgument::OPTIONAL, 'What environment should we use to build?', 'local')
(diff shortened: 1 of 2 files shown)


damiani:
@mattisbones Thanks so much for this contribution! 

I added `--no-warnings` to the snapshot tests so the prompt doesn't appear when running the test suite.

PR #281, opened 2018-10-07
Add interface PathResolverInterface and use it for type-hints.

Spudley:
Just been exploring the code, and felt it could use a bit more in the way of self-documenting type hints.

I started at SiteBuilder.php and the Path Resolvers. Haven't had time to do much more than this, but hopefully it is useful. It doesn't change anything, it just makes things easier to understand and work with.


JoshuaBehrens:
I added type hints and interfaces in my recent PR https://github.com/tightenco/jigsaw/pull/299. Can you review it whether this suits your case? Did I miss any spot whether to extract an interface from?

damiani:
This has now been superseded by #420.

PR #285, opened 2018-10-10
show drafts when config variable `drafts` is set to true, hide otherwise

### LINKED ISSUE #221: Hide draft posts from /build_production, but include in /build_local
Is there a way to hide/exclude a .md file/post from the posts collection when compiling to build_production, but have it included in build_local?

The YAML of the post.md file would be:
```md
---
extends: _layouts.post
title: post title
author: me
status: draft  // <-- this would change to status: published when ready, could also be draft: true/false
date: "2018-05-29"
---

lots of text.
```

I'm struggling with the collections helper methods. Thanks for the help.

[Proposal] Maybe add a "How to" section to the docs?

rumansaleem:
This resolves #221 by simply using a `drafts` boolean value in `config`.

config.php
```php
<?php

return [
    ...
    'drafts' => true,
    ...
];
```

config.production.php

```php
<?php

return [
    ...
    'drafts' => false,
    ...
];
```

So, now when you run `jigsaw build production`, drafts  inside `_drafts` subdirectory are _not_ included in collections and also _not_ compiled to static html. While, when you run `jigsaw build` or `jigsaw build local` all your drafts inside `_drafts` are include and compiled to static html.

Please review this, and let me know if I can improve this. I am not very good at testing, so please bare me with this if I did something stupid. :sweat_smile:


hellocosmin:
I wonder if this would be easier if it were based on a `date` Front Matter variable?

For example, what I always do in Laravel for drafts, is that instead of adding a `draft` or `published` column in the database, I work with the date directly. 

Here's an example scope on the model in Laravel:

```php
/**
 * Query scope for limiting the results to published posts only.
 *
 * @param Builder $query
 * @return Builder
 */
public function scopePublished(Builder $query)
{
    return $query->where('published_at', '<=', Carbon::now());
}
```

The `reject()` method could do something similar. This way, you'd only need to add a `date: YYYY-MM-DD` Front Matter key to your post/page, no need to create a folder with a name that is hardcoded in core. And you wouldn't need a `drafts` key either.

What do you think?

ruman63:
But that way, you'd need to build every next time to include the posts which are due for publishing. But still, a clean approach in my mind, you dont have to maintain `_drafts` folder.

rumansaleem:
Hey @hellocosmin, shall I make some changes to this PR? 
Maybe @damiani has something to say about this. We can discuss a little about how you really want to implement this feature of drafts or publishing. I'll be happy to help. :slightly_smiling_face:

rumansaleem:
I  have resolved the merge conflicts, and I guess its now ready to merge.

m1guelpf:
This PR assumes you're using Jigsaw in a specific way. I'd rather see this as some additional layer on top of Jigsaw instead of built-in.

stancl:
@m1guelpf is right, drafts seem very blog-specific. You wouldn't use them in, say, a docs project.

rumansaleem:
Hmm.. @m1guelpf you're right. This PR needs more work.

One of the nice approach seems to be the one suggested by @camuthig [here](https://github.com/tightenco/jigsaw/issues/221#issuecomment-478198703).
Is this the right way to go about it? I would like to know what people have to say regarding this approach. I'd start working in this direction if this one seems acceptable.

Is it wise to submit another PR and close this one, or just modify this one?

camuthig:
@rumansaleem I pushed the code that I was using to test locally to a branch https://github.com/camuthig/jigsaw/commit/b254fcce9158f80e677200b6a3b14596bd5d8ae1 if you want to take a look at what I was doing to make it work. Seeing as there has been a good deal of back and forth on how this functionality should be implemented, I thought it would be better to allow the discussion in the issue to finish prior to creating another PR that conflicted with this one.

rumansaleem:
thanks, I'll take a look.

damiani:
Addressed by https://github.com/tightenco/jigsaw/pull/379

PR #299, opened 2019-01-02
Adding type hints and interfaces

JoshuaBehrens:
To improve the ability to add own handlers or decorate them I added interfaces (contracts when speaking of laravel) to make clear what is needed for a new custom handler to implement. While doing this I had to get a feeling, what types are used and what they are used for. Therefore I added strict type hints for easier code completion and understanding of the data flow. This is a huge change in code internals but in the nature of php with dynamic typing|duck typing this should not affect existing jigsaw setups without any code changes. Any installations with custom handlers need to add type hints and implement the interfaces but these interfaces existed in a hidous way already.

This PR interferes with https://github.com/tightenco/jigsaw/pull/281 which gave me the idea for the PR. If I am referenced by name in any way @Spudley should be as well.


carbontwelve:
It's a shame this wasn't merged in. These hints and interfaces really help.

JoshuaBehrens:
Yikes I see all the conflicts

stevebauman:
I think this is a great PR, but it's definitely a bit opinionated. I think code docs and type-hints should at least be added as a minimum, rather than have return types for everything.

Just my two cents! :heart:

damiani:
This has now been superseded by #420.

Issue #302, opened 2019-01-03
Reconsider adding dotenv support


m1guelpf:
I remember seeing this proposal rejected in favor of environment config (`config.production.php` and the like), but I'd like this to be reconsidered for open-source.

While the current approach solves the problem of different configs on different build stages, it still leaks keys/secrets if you want to open-source your Jigsaw site (unless you set up complex deploy steps, like wget'ing a config file).

If you've changed your mind about this, I'd be more than happy to submit a PR :)

PR #310, opened 2019-01-10
Add dotenv support

m1guelpf:
Closes https://github.com/tightenco/jigsaw/issues/302.

Simple PR to add support for a `.env` file. You can directly use the `env()` function, as it's included in the `illuminate/support` package.

Let me know if I need to change anything, especially tests :)


m1guelpf:
ping @damiani :)

PR #324, opened 2019-02-21
rightTrim Directory Path

### LINKED ISSUE #322: writeOutputFile generates empty Folders in project root
Hi there 👋

This isn't necessarily an issue but more a little annoyance.
When I'm using `$jigsaw->writeOutputFile()` in a listener, Jigsaw creates an empty directory structure in the root of the project.

For example:
- My Jigsaw project is located here: `/Users/stefan/projects/example.org/`
- After running `jigsaw build` a new folder `Users` appears in the root of the project
- The new folder contains empty folders and matches the path of the project (eg. `/Users/stefan/projects/example.org/Users/stefan/projects/example.org/`)

I've seen, that the `GenerateIndex`-example in [tightenco/jigsaw-blog-template](https://github.com/tightenco/jigsaw-blog-template/blob/5a88893ff04db7e57846e3563c70bfcfbdec1cd7/listeners/GenerateIndex.php) uses `file_put_contents` instead of `$jigsaw->writeOutputFile()`. 
Is this on purpose? Or is this just to keep the example as simple as possible?

```php
<?php

namespace App\Listeners;

use TightenCo\Jigsaw\Jigsaw;

class GenerateIndex
{
    public function handle(Jigsaw $jigsaw)
    {
        $data = collect($jigsaw->getCollection('posts')->map(function ($page) use ($jigsaw) {
            return [
                'title' => $page->title,
                'categories' => $page->categories,
                'link' => rightTrimPath($jigsaw->getConfig('baseUrl')) . $page->getPath(),
                'snippet' => $page->getExcerpt(),
            ];
        })->values());

        // Using $jigsaw->writeOutputFile() would create empty folders
        // $jigsaw->writeOutputFile('index.json', json_encode($data));

        file_put_contents($jigsaw->getDestinationPath() . '/index.json', json_encode($data));
    }
}
```

I could pinpoint the cause of this bug to [this line](https://github.com/tightenco/jigsaw/blob/fb9cd18e36ec78a77d6151b37bbc41a6c12f4418/src/File/Filesystem.php#L22) in `TightenCo\Jigsaw\File\Filesystem`.

The value of `$file_path` in my example would be `/Users/stefan/projects/example.org/build_local/index.html`. 
But by calling `trimPath()` the first `/` get's removed (eg. `Users/stefan/projects/example.org/build_local`) and the folder structure get's created in the root of the project.

I guess `trimPath()` is there to fix compatibility issues between Linux, macOS and Windows? So just switching to `rightTrimPath()` would cause issues on Windows machines 🤔 

So my questions are: 

- Would you accept a PR which fixes this behavior?
- In which contexts should `writeOutputFile()` be used?

stefanzweifel:
Fixes #322.

The test would still work if `trimPath()` is used.
I just tried to test if `writeOutputFile` can handle a leading slash in `$file_path`.

### DIFF
--- a/src/File/Filesystem.php
+++ b/src/File/Filesystem.php
@@ -19,7 +19,7 @@ class Filesystem extends BaseFilesystem
     {
         $directory_path = collect(explode('/', $file_path));
         $directory_path->pop();
-        $directory_path = trimPath($directory_path->implode('/'));
+        $directory_path = rightTrimPath($directory_path->implode('/'));
 
         if (! $this->isDirectory($directory_path)) {
             $this->makeDirectory($directory_path, 0755, true);
--- a/tests/EventsTest.php
+++ b/tests/EventsTest.php
@@ -522,6 +522,17 @@ class EventsTest extends TestCase
 
         $this->assertEquals('test', $source->getChild('build/new_directory/file.html')->getContent());
     }
+
+    /** @test */
+    public function user_can_write_a_new_output_file_in_a_new_directory_in_after_build_event_listener_when_path_begins_with_slash()
+    {
+        $this->app['events']->afterBuild(function ($jigsaw) use (&$result) {
+            $result = $jigsaw->writeOutputFile('/new_directory/file.html', 'test');
+        });
+        $this->buildSite($source = $this->setupSource());
+
+        $this->assertEquals('test', $source->getChild('build/new_directory/file.html')->getContent());
+    }
 }
 
 class TestListener


stancl:
@stefanbauer Perhaps a regression test could be added, to make sure that no new non-`build_*` directory is created in the project's root after `jigsaw build` is ran?

damiani:
That was my plan, care to add it and then I'll merge?

stancl:
I'm unable to reproduce this issue in the test.

```php
/**
 * @test
 */
public function no_non_build_directories_are_created_after_the_project_is_built()
{
    $this->app['events']->afterBuild(function ($jigsaw) {
        $jigsaw->writeOutputFile('new_directory/file.html', 'test');
    });

    $vfs = $this->setupSource();

    $original_directory_structure = scandir($vfs->url());

    $this->buildSite($vfs);

    $new_directory_structure = scandir($vfs->url());

    $this->assertEquals(array_merge($original_directory_structure, ['build']), $new_directory_structure);
}
```

While this tests does fail due to different order of items in the compared arrays (this is an issue with the test), no new non-`build` directory gets created in the virtual file system.

stefanzweifel:
Quick update: I also tried to reproduce the problem with some tests, but when using vfsstream the issue doesn't show up.

I'm not familiar enough with the codebase, but maybe I can replace the code which builds the `$directory_path` and checks if the folder exists completely. 🤔 

I keep you posted 🤙

PR #325, opened 2019-02-22
Update Laravel dependencies to 5.7, replace Laravel helper calls

stancl:
Hi,

I changed the required versions of illiminate packages to 5.7 and replaced all global helper calls with respective Arr or Str calls, like https://github.com/laravel/framework/pull/26898. For future 5.9 support.

All tests pass so everything should be okay.


stancl:
I also reordered `use` statements so they're now in line with the way Laravel does it.

damiani:
The `Arr` and `Str` replacements look great, but for this project, we prefer to alphabetize `use` statements, unlike the Laravel core. (There's a PHP CS Fixer config in the repo, `.php_cs`, which I need to document in the readme.) If you can revert the two commits that relate to re-ordering `use` statements, I can merge this in. Thanks!

stancl:
Okay, reverted. Though I cannot guarantee that `use` statements introduced by me follow the alphabetic order. I'll see if I can perhaps run PHP CS Fixer on all files?

src/Collection/Collection.php
```diff
@@ -4,6 +4,7 @@
 
 use Closure;
 use Illuminate\Support\Collection as BaseCollection;
+use Illuminate\Support\Arr;
```
stancl: My `use` statements were simply appended to the end.

damiani:
Looks great, I'll run fixer after merge. Thanks for your contribution!

PR #347, opened 2019-06-07
Build dotfiles

o0-zzuf-0o:
The SiteBulider and tests use:
Illuminate\Filesystem\Filesystem allFiles()
instead of:
TightenCo\Jigsaw\File\Filesystem files()

As the second param ($hidden = false) isn't set to true, dotfiles are not built.

Added a .DS_Store into the test source

Added two explicit tests
* check that dotfile is present
* check that .DS_Store file isn't present

### DIFF
--- a/src/SiteBuilder.php
+++ b/src/SiteBuilder.php
@@ -73,7 +73,7 @@ class SiteBuilder
 
     private function generateFiles($source, $siteData)
     {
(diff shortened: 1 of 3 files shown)


damiani:
Great catch, thanks @o0-zzuf-0o !

PR #379, opened 2019-10-03
Support filtering collections

camuthig:
This PR creates the base functionality to allow filtering collections. Filtering allows for specific features, like not publishing draft messages, to be implemented by other projects.

Related to #221

### DIFF
--- a/src/Collection/Collection.php
+++ b/src/Collection/Collection.php
@@ -22,9 +22,12 @@ class Collection extends BaseCollection
 
     public function loadItems($items)
     {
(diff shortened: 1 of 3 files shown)


camuthig:
As implemented now, this PR only supports a single filter per collection. It wouldn't be much harder to support an array of filtering functions, though, if others have opinions.

cameron-elliott:
This seems like a cool feature!

damiani:
Love it...thanks @camuthig!

PR #392, opened 2019-11-07
Update Illuminate dependencies to 6.x

bakerkretzmar:
This pull request updates the `illuminate/container`, `illuminate/support`, and `illuminate/view` dependencies from `^5.8` to `^6.0`. That change in turn updates some other underlying dependencies, like `illuminate/event` and `nesbot/carbon`, but this doesn't appear to have any effect.

One test was using the old helper functions, so I updated those from `array_get` to `Arr::get`. All tests that were passing before this change are still passing (`MarkdownExtraTest::correctly_parse_single_line_html_markup_in_markdown_file` and `SnapshotTest::all_built_files_contain_expected_content` were both already failing).


batFormat:
Please merge this PR.

damiani:
Thanks for this @bakerkretzmar! This got a little stale after some recent merges, so I opened https://github.com/tightenco/jigsaw/pull/421 to reconcile matters.

PR #416, opened 2020-01-30
Use Markdown extra instead of Parsedown extra

Jampire:
Hi. Is there any reason to use Parsedown Extra instead of Markdown Extra? Parsedown Extra declares itself as '_an extension that adds support for Markdown Extra_'. But in fact, it doen't. There some items that still are not implemented in Parsedown Extra. So, there is a reason to switch to full support of extra markdown - switch to [Markdown Extra](https://github.com/michelf/php-markdown). Here is my PR. I just added the wrapper to generate Table of Contents (TOC). You can read [here](https://dzianiskotau.com/blog/php/auto-generating-toc-markdown/) how it can be used.


damiani:
I'd definitely be up to switching to `michelf/php-markdown` if it doesn't break anything—an initial run through of the tests seems to indicate that it would be a clean swap, with the exception of slightly different line break behavior.

I'm not sold on adding the TOC/automatic ID generation by default, though, which feels out of scope for the parser. Users could, however, swap or extend the parser in `bootstrap.php` if they wanted to use `jampire/toc-markdown-extra`.

Jampire:
I agree that TOC generation is the choice of the user and should be configured in the application. Just a note here: `michelf/php-markdown` has bult-in capabilities to use user-defined functions instead of its own ones. For example, it already has the ability to generate headers ids. You just need to declare function for ids generating. There are also some other functions that you can declare or override.

damiani:
Great, I made those updates. Thanks for this @Jampire !

Issue #424, opened 2020-02-13
Code block CSS class missing language prefix


jreviews:
In a previous version when I used a code block with a specific language, the class name would appear as `language-php`. After updating to the latest version, the class name is now only `php`.

Is there a configuration setting somewhere to be able to automatically add the language prefix?

Or should the code blocks now be marked up to also include the prefix?

![image](https://user-images.githubusercontent.com/31069418/74439927-78855200-4e43-11ea-9c68-8e2852554898.png)

damiani:
The latest version switched Markdown parsers, so I suspect that's the source of this change, but I'll take a look!

Jampire:
Both `php` and `language-php` are fine here. Code highlighting is done by `highlight.js` or similar package. Markdown parser just inserts the class that you assigned to it. In this case it's `php` or `language-php`. Please, show your final html parsed code and check what package you are using to highlight the code.

jreviews:
I understand that both are fine. The issue is that before `php` alone would result in a final html parsed code with `class="language-php"` and after the change in Markdown parsers  the output is just `class="php"`.

Jampire:
@jreviews are you using Prism for syntax highlight? Prism requires `language-*` class to be added to class attribute. That's why Parsedown generates it. But highlight.js, shipped with Jigsaw templates, doesn't require it. If you are not using highlight.js, could you try it as workaround?

jreviews:
I can make changes needed inside my project to get this working. However, the point of this issue is that any previous project that expected to have the `language-*` prefix as it did before, now needs to be updated to take this into account.

damiani:
@jreviews Fortunately, the new parser exposes a configuration setting for that class name prefix, and I think we should set the default to match the previous Parsedown behavior so we don't break a bunch of existing Jigsaw sites. I've updated that in #426, and also added a way for those config settings to be more easily accessed in `bootstrap.php`, should anyone want to remove the `language-*` prefix (or change any other [PHP Markdown setting](https://michelf.ca/projects/php-markdown/configuration/)).

jreviews:
Thank you!

PR #426, opened 2020-02-24
Restore "language-" prefix for fenced code blocks

damiani:
As referenced in https://github.com/tightenco/jigsaw/issues/424, the recent change in Markdown parsers from https://github.com/tightenco/jigsaw/pull/416 altered the default behavior for language class names applied to fenced code blocks. The previous parser (Parsedown) followed the Prism convention of prefixing language class names with `language-*`, whereas the new parser (PHP Markdown) omits the prefix.

This PR restores the previous Jigsaw behavior, to prevent breaking sites that relied on the `language-` prefix for syntax highlighting. 

In addition, since PHP Markdown exposes this setting (and others) through a series of configuration options, this PR makes those settings easily configurable in `bootstrap.php`. Specifically:

- it updates the MarkdownParser binding to use a singleton, and 
- binds the parser to a `markdownParser` key on the container

Now, any of the parser's [configuration options](https://michelf.ca/projects/php-markdown/configuration/) can be easily set in `bootstrap.php`. For example, to remove the `language-` prefix from fenced code block class names, this line can now be added to `bootstrap.php`:

```php
$container['markdownParser']->code_class_prefix = '';
```

Likewise, this PR will make it possible to include a custom function for generating IDs (as mentioned [here](https://github.com/tightenco/jigsaw/pull/416#issuecomment-580306406)).

### DIFF
--- a/jigsaw-core.php
+++ b/jigsaw-core.php
@@ -84,7 +84,13 @@ $container->bind('outputPathResolver', function ($c) {
 
 $container->bind(YAMLParser::class, SymfonyYAMLParser::class);
 
-$container->bind(FrontYAMLMarkdownParser::class, MarkdownParser::class);
+$container->singleton('markdownParser', function ($c) {
+    return new MarkdownParser;
+});
+
+$container->bind(FrontYAMLMarkdownParser::class, function ($c) {
+    return $c['markdownParser'];
+});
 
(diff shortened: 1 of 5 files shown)


jreviews:
Works great, but instead of hyphen you used an underscore 877fa65168f38ce35d86fcc6293a2db276edc216

Issue #400, opened 2019-12-24
Collection custom path


huguesjoyal:
Hi,

It would be great to be able to access relativePath or the file object in the collection custom closure.

So we could do something like :

```php
'docs' => [
    'path' => function($page) {
        return $page->getCollection().'/'.$page->getRelativePath().'/'.$page->getFilename();
     }
],

'docs' => [
    'path' => '{collection}/{relativePath}/{-filename}',
```

If accepted I could do the following merge request in the file CollectionDataLoader.php
```php
    private function getMetaData($file, $collection, $data)
    {
        $filename = $file->getFilenameWithoutExtension();
        $baseUrl = $data->baseUrl;
        $extension = $file->getFullExtension();
        $collectionName = $collection->name;
        $collection = $collectionName;
        $source = $file->getPath();
        $modifiedTime = $file->getLastModifiedTime();
        $relativePath = $file->getRelativePath();

        return compact('file', 'filename', 'baseUrl', 'extension', 'collection', 'collectionName', 'source', 'modifiedTime', 'relativePath');
    }
```

damiani:
Yes, great idea; feel free to PR the addition of `$file->getRelativePath()` to `getMetaData()` 👍🏻

PR #428, opened 2020-02-27
Add relativePath to collection issues #400

huguesjoyal:


damiani:
Great! I think this should also be added to `getMetaData()` in `SiteBuilder.php`, for files that aren't part of a collection.

huguesjoyal:
Just added... I didn't used it there for my work... but still a good idea.

PR #429, opened 2020-02-29
Expose source file info to afterBuild event

damiani:
This PR adds a `getSourceFileInfo()` method to the Jigsaw class, for reference during the `afterBuild` event. 

`getSourceFileInfo()` returns a collection of all output files that were generated. For each item:
- the key contains the path of the output file relative to the `build` directory (e.g. `/posts/my-first-post`)
- the value contains the Jigsaw `InputFile` object for the source file. On `InputFile`, every method from PHP's [SplFileInfo](https://www.php.net/manual/en/class.splfileinfo.php) class is available.

This is similar to `getOutputPaths()`, which returns just the output paths (i.e., the keys from `getSourceFileInfo()`).

### DIFF
--- a/src/File/CopyFile.php
+++ b/src/File/CopyFile.php
@@ -6,10 +6,10 @@ class CopyFile extends OutputFile
 {
     protected $source;
 
-    public function __construct($source, $path, $name, $extension, $data, $page = 1)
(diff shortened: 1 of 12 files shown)


damiani:
Note: this has been replaced by #433.

Issue #289, opened 2018-11-02
Jigsaw doesn't remove empty temporary directories


Anthony-Gaudino:
When using remote collections Jigsaw will create temporary directories in the format name `_XXXX` and then move the resulting files from this directory to elsewhere.

I suggest having Jigsaw look for any empty  directory starting with an underscore ("_") after building the website and then removing them.

anthonyterrell:
Hey @Anthony-Gaudino , 

The default behavior of the build process is to remove the temporary directories. If there are still left over `_xxx` directories, it is likely another error has occurred. 

If you're still running into this, please let us know a little bit about the error you're running into. We'd be more than happy to put our heads together and figure it out 😃

anonlethal-jerk:
I know this was closed, but I ran into this issue as well; source folder littered with empty `_collection` directories after successful builds. The `_tmp` directories were being cleaned, but not their parents.

I was able to fix this by changing the cleanup function in `CollectionRemoteItemLoader.php` to the following:

```
    public function cleanup()
    {
        collect($this->tempDirectories)->each(function ($path) {
            $this->files->deleteDirectory($path);

            $tempParent = $this->files->dirname($path);
            if ( $this->files->isDirectory($tempParent) ) {
                // Get all files in this directory.
                $files = $this->files->files($tempParent);
                // Check if directory is empty.
                if (empty($files)) {
                    // Yes, delete the directory.
                    $this->files->deleteDirectory($tempParent);
                }

              }
        });
    }
```

I am not a pro-coder though, so YMMV on if this is good/best way to do so.

damiani:
Right you are, thanks @anonlethal-jerk !

PR #435, opened 2020-03-08
Remove empty temp remote collection folders after build

anonlethal-jerk:
I use a number of remote collections and after builds (even successful builds), the `source` folder is filled with all the newly created empty `_collection_name` folders. Looks like cleanup routine clears the `_tmp` folders created, but not their empty parents.

Related to issue #289


damiani:
Super! I modified the code a bit to use the `isEmptyDirectory` method, and updated/added a test.

Great work, thanks @anonlethal-jerk !

PR #447, opened 2020-04-05
Added support for mapping over collection inside of config

ryangjchandler:
As with the `filter` config option, this pull request adds support for mapping over the items after filtering.

It would be done like this:

```php
return [
    'collections' => [
        'posts' => [
            'map' => function ($post) {
                return new Post($post);
            }
        ],
    ]
]
```

The example above would let you map the collection to a data object for example, where you could place larger helper methods.

If not sure which way round makes more sense, whether the items should be mapped and then filtered, or filtered then mapped. Mapping after makes sense since it might be a longer mapping process / more resource consumption.

### DIFF
--- a/src/Collection/Collection.php
+++ b/src/Collection/Collection.php
@@ -26,6 +26,7 @@ class Collection extends BaseCollection
         $sortedItems = $this
             ->defaultSort($items)
             ->filter($this->getFilter())
+            ->map($this->getMap())
             ->keyBy(function ($item) {
                 return $item->getFilename();
             });
@@ -64,6 +65,19 @@ class Collection extends BaseCollection
 
         return function ($item) {
             return true;
+        };
+    }
+
+    private function getMap()
+    {
+        $map = Arr::get($this->settings, 'map');
+
+        if ($map) {
+            return $map;
+        }
+
+        return function ($item) {
+            return $item;
         };
     }
 
--- a/src/Collection/Collection.php
+++ b/src/Collection/Collection.php
@@ -25,8 +25,8 @@ class Collection extends BaseCollection
     {
         $sortedItems = $this
             ->defaultSort($items)
-            ->filter($this->getFilter())
             ->map($this->getMap())
+            ->filter($this->getFilter())
             ->keyBy(function ($item) {
                 return $item->getFilename();
             });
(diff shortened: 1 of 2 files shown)


mattstauffer:
This is definitely an interesting idea! Thanks for the PR-- @damiani will take a look soon. Thanks so much Ryan!

damiani:
I think this is a fantastic idea, and mapping after filtering (as you have it in the PR) makes the most sense to me. I'm curious: what are you doing in your `Post` class?

ryangjchandler:
Pseudo relationships in the class, for example with my webmentions. Each webmention gets picked up and transformed into its own Webmention class. They're basically like models, without the database.

damiani:
@ryangjchandler On second thought, filtering after mapping is probably a better idea, so the filtering could take advantage of helper methods in the new item class, or the class can set item variables before hitting the filter.

ryangjchandler:
Yeah that is one benefit, but for larger collections you would need to map over every item before filtering which might drastically affect build time. Either way is fine by me!

ryangjchandler:
Oh yeah, I forgot test :/

damiani:
I think in most cases, the impact of instantiating a new class would be minimal; this is an advanced feature anyway, so we'll trust people to know what they're doing :)

ryangjchandler:
> I think in most cases, the impact of instantiating a new class would be minimal; this is an advanced feature anyway, so we'll trust people to know what they're doing :)

For sure, I'll change that around now :)

damiani:
Very excited about this, thanks for the contribution @ryangjchandler !

ryangjchandler:
> Very excited about this, thanks for the contribution @ryangjchandler !

Anytime! Would be great to discuss any larger ideas you've got :)

lokmanm:
@damiani this is what I was thinking about my View Models feature idea, I'm very happy someone did it. Thanks a lot @ryangjchandler !

ryangjchandler:
> @damiani this is what I was thinking about my View Models feature idea, I'm very happy someone did it. Thanks a lot @ryangjchandler ! 
> 
> 

Glad to here it. I've got a couple of ideas for some higher level abstraction improvements for the project that I'll definitely benefit from.

PR #454, opened 2020-04-11
use Process::fromShellCommandline instead of constructor

faustbrian:
> ./vendor/bin/jigsaw init docs

`PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Symfony\Component\Process\Process::__construct() must be of the type array, string given, called in vendor/tightenco/jigsaw/src/Scaffold/ProcessRunner.php`


damiani:
Thanks for this fix @faustbrian

PR #458, opened 2020-04-20
Fix .env files due to breaking change on vlucas/phpdotenv 4.0

keithbrink:
The latest upgrade to vlucas/phpdotenv 4.0 broke using .env files, because of a breaking change made from 3 -> 4: https://github.com/vlucas/phpdotenv/blob/master/UPGRADING.md#v3-to-v4

This fixes the issue and changes the test to use an actual .env file.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+JIGSAW_TEST_VAR=true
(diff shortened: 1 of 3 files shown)


damiani:
Thanks @keithbrink !

keithbrink:
Thanks for the quick merge and tagging a release!

Issue #488, opened 2020-07-28
Build error on .git file in collection folder


majorminors:
## expected behaviour
If `.git` is present in a collection folder, the build process ignores it.

## actual behaviour
If `.git` is present in a collection folder, the build process does not finish.

## detail and notes
I've abstracted my articles to a separate git repo, and pulled them into the jigsaw project as a git submodule. I just noticed that the local build seems to error and I am assuming this is because it doesn't like the `.git` in the `source/_articles` directory. See the printout from npm run watch:
```
 DONE  Compiled successfully in 14269ms                                                                                                                                               21:48:00

       Asset      Size    Chunks             Chunk Names
 /js/main.js  1.31 MiB  /js/main  [emitted]  /js/main
css/main.css  1.45 MiB  /js/main  [emitted]  /js/main

In CollectionDataLoader.php line 104:
                                                       
  No matching collection item handler for file: .git.  
                                                       

build [--pretty PRETTY] [-c|--cache [CACHE]] [--] [<env>]


[Browsersync] Access URLs:
 -------------------------------------
       Local: http://localhost:3000
    External: http://192.168.0.23:3000
 -------------------------------------
          UI: http://localhost:3001
 UI External: http://localhost:3001
 -------------------------------------
[Browsersync] Serving files from: build_local/
```

Assuming this isn't something I've broken, let me know if there's anything I can do to help troubleshoot.

Warm regards,

Dorian

bakerkretzmar:
Interesting, I don't know if we've ever come across this use case before. It should be a straightforward fix, we're already filtering out files that start with `_` so just checking for `.` too will probably do it. I'll take a closer look this week.

majorminors:
Oh fabulous. I looked myself to see if I could fix it, but I'm not familiar with PHP, so I couldn't make much out of the `CollectionDataLoader.php` file. Apologies.

bakerkretzmar:
Actually, can you try [filtering](https://jigsaw.tighten.co/docs/collections-filtering/) out `.git` directly? Something like this:

```php
'filter' => function ($item) {
    return $item->filename != '.git';
}
```

majorminors:
Same error.

Although, just noticing how I'm dealing with the filename parameter for path:
`'path' => '{collection}/{filename}',`
Do I need to wrap it?

bakerkretzmar:
[...]

PR #494, opened 2020-08-07
add filter to ignore '.git' in CollectionDataLoader.php

majorminors:
See issue [488](https://github.com/tightenco/jigsaw/issues/488).

In `CollectionDataLoader.php`, appended:
Line 63: `return Str::startsWith($file->getFilename(), '_');`
with filter:
Line 63: `return Str::startsWith($file->getFilename(), '_') || Str::startsWith($file->getFilename(), '.git');`

### DIFF
--- a/src/Loaders/CollectionDataLoader.php
+++ b/src/Loaders/CollectionDataLoader.php
@@ -60,7 +60,7 @@ class CollectionDataLoader
 
         return collect($this->filesystem->files($path))
             ->reject(function ($file) {
-                return Str::startsWith($file->getFilename(), '_') || Str::startsWith($file->getFilename(), '.git');
+                return Str::startsWith($file->getFilename(), ['_', '.git']);
             })->tap(function ($files) {
                 $this->consoleOutput->progressBar('collections')->addSteps($files->count());
             })->map(function ($file) {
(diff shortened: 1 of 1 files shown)


majorminors:
May not suit your needs, but this is working for me and seems to be in line with the thinking of @bakerkretzmar , so thought I'd share. Thanks to @joshhanley.

src/Loaders/CollectionDataLoader.php
```diff
@@ -60,7 +60,7 @@ private function buildCollection($collection)
 
         return collect($this->filesystem->files($path))
             ->reject(function ($file) {
-                return Str::startsWith($file->getFilename(), '_');
+                return Str::startsWith($file->getFilename(), '_') || Str::startsWith($file->getFilename(), '.git');
```
bakerkretzmar, line 63:
```suggestion
                return Str::startsWith($file->getFilename(), ['_', '.git']);
```

bakerkretzmar:
Thanks! Just want to make sure this isn't a breaking change, but then looks good to me!

src/Loaders/CollectionDataLoader.php
joshhanley: TIL. I didn't know you could pass an array into the starts with function. That's much cleaner!

joshhanley:
@bakerkretzmar I can't imagine it would break anything. The only other consideration would be to do an exact match instead of a starts with. That would ensure it only filters out .git files.

But then I just did a test if you had a .gitignore inside a collections directory, it breaks in the same way "No matching collection item handler for file: .gitignore", so maybe leaving it as starts with .git would be best.

PR #495, opened 2020-08-10
Windows compatibility

royvanv:
This PR improves the compatibility with Windows.

Not really sure if commit 9c21733 is the best solutions.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -98,13 +98,8 @@ class TestCase extends BaseTestCase
         return str_replace("\n", "", $output);
     }
 
-    protected function fixDirSlashes(string $path): string
+    protected function fixDirectorySlashes(string $path): string
     {
-        if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN')
-        {
-            return str_replace('/', '\\', $path);
-        }
-
-        return $path;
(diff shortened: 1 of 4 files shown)


tests/TestCase.php
```diff
@@ -97,4 +97,14 @@ public function clean($output)
...
     }
+
+    protected function fixDirSlashes(string $path): string
+    {
+        if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN')
+        {
+            return str_replace('/', '\\', $path);
+        }
+
+        return $path;
```
damiani, line 108: I believe we can remove the conditional and simplify this to:

```php
return str_replace(['/', '\\'], DIRECTORY_SEPARATOR, $path)
```

tests/TestCase.php
```diff
@@ -97,4 +97,14 @@ public function clean($output)
     {
         return str_replace("\n", "", $output);
     }
+
+    protected function fixDirSlashes(string $path): string
```
damiani, line 101: Can you rename this to `fixDirectorySlashes`, or, if that's too long, something like `fixSlashes` or `fixPath`? We prefer to not use shortened words wherever possible.

damiani:
Thanks for this! I left a couple comments in `TestCase`.

damiani:
Thanks @royvanv!

PR #497, opened 2020-08-19
Prefix test build commands with `php` for Windows compatibility.

royvanv:
PHPUnit setup was not able to execute `./jigsaw` directly. Prefixing it with `php` fixes this issue on Windows.

Please test this on macOS and Linux.


damiani:
Thanks @royvanv!

Issue #492, opened 2020-08-05
Add support for Blade component classes


royvanv:
Please add support for component classes for [Blade components](https://laravel.com/docs/7.x/blade#components).

bakerkretzmar:
You can register class-based Blade components in Jigsaw with `$bladeCompiler->component()`. See #440.

royvanv:
This basic example in my `blade.php` does not work. I have not created a view for it yet. I use `<x-icon />` in my view.

```php 
use Illuminate\View\Component;

class Icon extends Component
{
    public function render()
    {
        return "Icon works!";
    }
}

$bladeCompiler->component(Icon::class);
```

This code results in an error when building the site:

> In 05187dc394c0a6d7c726c66dad2543a3097e1c8e.php line 22:
> 
>   syntax error, unexpected 'endif' (T_ENDIF), expecting end of file (View: [...]\source\_layouts\master.blade.php) (View: [...]  
>   \source\_layouts\master.blade.php)
> 
> 
> In 05187dc394c0a6d7c726c66dad2543a3097e1c8e.php line 22:
> 
>   syntax error, unexpected 'endif' (T_ENDIF), expecting end of file (View: [...]\source\_layouts\master.blade.php)
> 
> 
> In 05187dc394c0a6d7c726c66dad2543a3097e1c8e.php line 22:
> 
>   syntax error, unexpected 'endif' (T_ENDIF), expecting end of file
> 
> 
> build [--pretty PRETTY] [-c|--cache [CACHE]] [--] [<env>]

It creates the following compiled code in the cache file:
```php
<?php $__env->startComponent('Example', 'example', []); ?>
<?php $component->withAttributes([]); ?>
<?php if (isset($__componentOriginal0f01ed56a1e32a05e5ef96e4d779f34784af9a96)): ?>
<?php $component = $__componentOriginal0f01ed56a1e32a05e5ef96e4d779f34784af9a96; ?>
<?php unset($__componentOriginal0f01ed56a1e32a05e5ef96e4d779f34784af9a96); ?>
<?php endif; ?>
<?php echo $__env->renderComponent(); ?>
<?php endif; ?>
```

bakerkretzmar:
@royvanv sorry for closing this prematurely, you're right, that example doesn't work. Forcing the classname of the component you're registering to include a `\` appears to fix this, so for now you can:

[...]

PR #507, opened 2020-10-22
Support autoloading class components

jobyh:
Ahoy-hoy splendid people of Tighten thanks for Jigsaw it's rad-biscuits 💜

I'm rebuilding a static site (originally Gatsby) and have been finding the manual registration of class-based Blade components as discussed in feels #492 a bit of a fiddle / doesn't scale well. For context I'm using inline Blade class components pretty much exclusively (_love_ those inline class components 🙌).

This PR builds on #440 proposes a solution for autoloading based on a `Components` namespace being used. I have intentionally left adding this mapping to `composer.json` as a manual step. After looking at adding this programmatically via `init` I found ensuring correct integration with subclasses of `ScaffoldBuilder` inluding 3rd party presets started throwing up too many questions and my internal KISS 🚨 alarm sounded. A one-off manual opt-in solves the problem at hand and doesn't feel too much like a kludge to me.

Here's the proposed usage:
```
% composer require tightenco/jigsaw
% ./vendor/bin/jigsaw init
```

... add the following to `composer.json`:

```
"autoload": {
  "psr-4": {
    "Components\\": "where/you/want/to/keep/component/classes/"
  }
}
```

...lastly

```
% composer dump-autoload
```

After that any component classes using the `Components` namespace will be automatically discovered when used via the `x-component` syntax in Blade files.

Couple of outstanding questions in my mind which it would be good to hear your thoughts on 🤔

- Should there be a namespace prefix to prevent collisions? (I felt this was overkill for Jigsaw)
- Not super happy about using `class_alias` in the tests due to polluting the test environment but finding a way to usefully mock seemed really convoluted at best. I have mitigated by aliasing to class names which are different to the fixture `Alert` classes in the test file but if you can think of a better approach would be great to hear

Thanks!

### DIFF
--- a/src/View/ComponentTagCompiler.php
+++ b/src/View/ComponentTagCompiler.php
@@ -4,6 +4,7 @@ namespace TightenCo\Jigsaw\View;
 
 use Exception;
 use Illuminate\Container\Container;
+use Illuminate\Support\Str;
 use Illuminate\View\Compilers\ComponentTagCompiler as BaseComponentTagCompiler;
 use Illuminate\View\Factory;
 
@@ -28,8 +29,20 @@ class ComponentTagCompiler extends BaseComponentTagCompiler
             return $view;
         }
 
+        if (class_exists($class = $this->guessClassName($component))) {
+            return $class;
+        }
+
         throw new Exception(
             "Unable to locate a class or view for component [{$component}]."
         );
+    }
+
+    public function guessClassName(string $component)
+    {
+        $componentPieces = array_map(function ($componentPiece) {
+            return ucfirst(Str::camel($componentPiece));
+        }, explode('.', $component));
+        return 'Components\\'.implode('\\', $componentPieces);
     }
 }
--- a/tests/BladeComponentTest.php
+++ b/tests/BladeComponentTest.php
@@ -116,6 +116,55 @@ class BladeComponentTest extends TestCase
             $built
         );
     }
+
+    /**
+     * @test
+     */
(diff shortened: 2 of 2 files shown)


jobyh:
Just found a possible bug using this locally. Closing PR while I investigate.

jobyh:
Some issues using `view()` under the `Components` namespace which the added tests didn't catch and some attention needed around mix `watch` not reloading. I'll reopen when have looked at these.

jobyh:
Ok we're good to go again:

In terms of `view()` the added tests are correct and the issue was in the meatware manually testing 🤪 - views should be returned via the dependency injection container as illustrated by the [existing test fixture](https://github.com/tighten/jigsaw/blob/58eae8ae2ec302bbd972187066d273e530f91778/tests/BladeComponentTest.php#L134) - `view()` helper isn't available in Jigsaw.

For `mix` to reload when updates are made to the component class directory (configured in `composer.json`) I was able to override the watch paths configuration as per the [Jigsaw Mix plugin README](https://github.com/tighten/laravel-mix-jigsaw/blob/main/README.md):

```
mix.jigsaw({
  open: false,
  watch: [
    'config.php',
    'source/**/*.md',
    'source/**/*.php',
    'source/**/*.scss',
    '!source/**/cache/*',
    'components/**/*.php',
    '!source/**/*.swp',
  ],
}
```

...I've been continuing to use this setup with my fork without further issue over the last week and a bit.

bakerkretzmar (approved):
@jobyh this is awesome, thank you so much!

damiani:
@jobyh Love it, thanks for this!

PR #510, opened 2020-10-27
Use Tailwind CSS and PostCSS instead of SCSS in default installation

jasonlbeggs:
This PR makes the necessary changes to include Tailwind CSS and PostCSS in the default installation instead of SCSS.

- Updated NPM dependencies
- Updated `webpack.mix.js` configuration
- Added `tailwind.config.js` with the `purge` option configured
- Added some very basic styles to the master layout and index page

### DIFF
--- a/stubs/mix/tailwind.config.js
+++ b/stubs/mix/tailwind.config.js
@@ -6,7 +6,6 @@ module.exports = {
       'source/**/*.html',
     ]
   },
-  darkMode: false,
   theme: {
     extend: {},
   },
--- a/stubs/mix/tailwind.config.js
+++ b/stubs/mix/tailwind.config.js
@@ -1,0 +1,17 @@
+module.exports = {
+  purge: {
+    content: [
+      'source/**/*.blade.php',
+      'source/**/*.md',
+      'source/**/*.html',
+    ]
+  },
+  darkMode: false,
+  theme: {
+    extend: {},
+  },
+  variants: {
+    extend: {},
+  },
+  plugins: [],
+}
--- a/stubs/site/source/_assets/css/main.css
+++ b/stubs/site/source/_assets/css/main.css
@@ -1,0 +1,3 @@
+@import 'tailwindcss/base';
+@import 'tailwindcss/components';
+@import 'tailwindcss/utilities';
--- a/stubs/mix/package.json
+++ b/stubs/mix/package.json
@@ -10,9 +10,9 @@
     },
     "devDependencies": {
         "cross-env": "^3.2.3",
-        "laravel-mix": "^4.0.0",
+        "laravel-mix": "^5.0.5",
         "laravel-mix-jigsaw": "^1.1.0",
-        "sass": "^1.15.2",
-        "sass-loader": "^7.1.0"
+        "postcss-import": "^12.0.0",
+        "tailwindcss": "^1.9.6"
     }
 }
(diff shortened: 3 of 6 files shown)


stubs/mix/tailwind.config.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+  purge: {
+    content: [
+      'source/**/*.blade.php',
+      'source/**/*.md',
+      'source/**/*.html',
+    ]
+  },
+  darkMode: false,
```
bakerkretzmar, line 9: `false` is the default, right? I can see the value of including the other empty default values here like for `theme` and `variants`, but idk if this one is necessary.

stubs/site/source/_assets/css/main.css
```diff
@@ -0,0 +1,3 @@
+@import 'tailwindcss/base';
+@import 'tailwindcss/components';
+@import 'tailwindcss/utilities';
```
bakerkretzmar, line 3:
```suggestion
@tailwind base;
@tailwind components;
@tailwind utilities;
```

This would let us remove the `postcss-import` plugin and dependency. Thoughts?

With `postcss-import` if you want to add custom CSS you _have_ to put it in another file and `@import` that file, this way you can add it to `main.css`.

bakerkretzmar (approved):
Heck yeah @jasonlbeggs 🙌🏻  thanks so much! About time we got rid of Sass 😄 

Looks great to me, I left a couple inconsequential comments for you and @damiani to make the call on.

jasonlbeggs:
@bakerkretzmar that's what I thought as well! I talked to @damiani about it before doing any work, and he's on board too!

stubs/site/source/_assets/css/main.css
jasonlbeggs: @damiani and I spoke about this one earlier. Neither one of us really use postcss-import, but the thought was maybe it's make the transition from scans a little easier? I'm pretty neutral on including it or removing it.

stubs/mix/tailwind.config.js
bakerkretzmar: Oh yeah sorry I only meant that one line, just `darkMode: false`

jasonlbeggs: Oh gotcha! Yeah we can remove that. I just pulled the default stub from Tailwind and configured purge.

stubs/site/source/_assets/css/main.css
damiani: Let's keep it in for now; I do think it will ease the transition for people used to using @imports from Sass. We can revisit at some point down the road.

stubs/mix/tailwind.config.js
damiani: Agreed, I removed it.

damiani:
Thanks so much for this @jasonlbeggs !

raniesantos:
Stumbled upon this while browsing changes.

Isn't Tailwind too specific of a tool? Sass is more generalized and basically just improved CSS. Tailwind is (in a way) a framework which enforces very specific ways of doing things.

Can I get clarifications on how you reached this decision???

damiani:
@raniesantos We love Tailwind in part because it helps users get up and running with a professionally-styled site very quickly. But, you're free to use whatever preprocessor (or none) and library (or none) that you'd like! We've documented [how to switch preprocessors](https://jigsaw.tighten.co/docs/compiling-assets/#enabling-different-preprocessors) if you want to use something else.

PR #520, opened 2020-12-16
PHP 8

c-fitzmaurice:
This PR bring `php8` and `phpunit:9.3` compatibility to Jigsaw, bumping the version to `1.3.33`.


src/Scaffold/PresetPackage.php
```diff
@@ -59,6 +60,8 @@ public function runInstaller($console)
             throw $e;
         } catch (Exception $e) {
             throw new Exception("The 'init.php' file for this preset contains errors.");
+        } catch (Error $e) {
+            throw new Exception("The 'init.php' file for this preset contains errors.");
```
bakerkretzmar, line 64: Is this specific to PHP 8?

bakerkretzmar:
With a few tweaks to the GitHub Actions workflow this works great. See bakerkretzmar/jigsaw#1.

@damiani I had to remove `composer.lock` to get the installs to work properly, there might be another way but we could also just remove it from the repo? My understanding is it's important to keep lockfiles around for apps, but better to omit them for packages and libraries.

src/Scaffold/PresetPackage.php
c-fitzmaurice: I'm not 100% certain, but in PHP 8 it was throwing an `Error` specific to improper php formatting and not the expecting `Exception` as it was before.

c-fitzmaurice:
@bakerkretzmar I pulled in the changes you made to the GitHub Action workflow and removed the `composer.lock` file here as well.

Let me know if this needs anything else!

PR #530, opened 2021-01-20
Update scaffold dependencies and formatting

bakerkretzmar:
Updates the bare-bones site scaffold included with Jigsaw to use Laravel Mix 6 and Tailwind 2, and:

- Uses the new Mix CLI to clean up the npm scripts.
- Removes the dependency on `cross-env`.
- Bumps the `laravel-mix-jigsaw` dependency to `^1.5.0` and `postcss-import` to the most recent version.
- Explicitly requires `postcss` in `package.json` since Mix installs it on the first run anyway.
- Adds the generated JS asset to the main layout.
- Removes `shrink-to-fit=no` from the viewport meta tag, it's specific to Safari 9.0.0-9.2.x. ([source](https://www.scottohara.me/blog/2018/12/11/shrink-to-fit.html))
- Removes the `x-ua-compatible` meta tag, it's basically only useful for IE 8-10. ([source](https://stackoverflow.com/a/6771584/6484459))
- Removes the `referrer` meta tag, `always` isn't a valid value for it and the default, `no-referrer-when-downgrade`, is good. ([source](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#directives))
- Renames the base layout from `master.blade.php` to `main.blade.php`.
- Moves all the stub files into `stubs/site` instead of keeping mix/npm/tailwind stuff separate.


damiani:
Love this! What about `main` for the primary template instead of `site`?

PR #541, opened 2021-02-05
Bind page data into container

bakerkretzmar:
This PR binds Jigsaw's page data into the container as `'pageData'` immediately before rendering each Blade template, making it possible to inject into components as described in https://github.com/tighten/jigsaw/issues/492#issuecomment-762669802:

```php
<?php

namespace Components;

use Illuminate\View\Component;
use Illuminate\Container\Container;

class Layout extends Component
{
    public $page;

    public function __construct()
    {
        $this->page = Container::getInstance()->make('pageData');
    }

    public function render()
    {        
        return Container::getInstance()->make('view')->make('_components.layout');
    }
}
```

```blade
{{-- source/index.blade.php --}}

{{-- No longer necessary to pass $page in manually! --}}
<x-layout>
    <div class="p-8">
        <h1 class="text-3xl font-bold">Hello world!</h1>
    </div>
</x-layout>
```

### DIFF
--- a/src/SiteBuilder.php
+++ b/src/SiteBuilder.php
@@ -2,9 +2,10 @@
 
 namespace TightenCo\Jigsaw;
 
-use TightenCo\Jigsaw\File\InputFile;
-use TightenCo\Jigsaw\File\Filesystem;
+use Illuminate\Container\Container;
 use TightenCo\Jigsaw\Console\ConsoleOutput;
+use TightenCo\Jigsaw\File\Filesystem;
+use TightenCo\Jigsaw\File\InputFile;
 
 class SiteBuilder
 {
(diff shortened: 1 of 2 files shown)


damiani:
What do you think about either (a) adding a `UsesPageData` trait to encapsulate this, and make it easier for a user to add to a component; or (b) adding a Jigsaw `Component` that extends `Illuminate\View\Component`?

(a) would require a constructor in a trait, which is a little iffy, but luckily `Component` doesn't have its own.

Thoughts?

bakerkretzmar:
@damiani a trait would be handy, but I also don't like the idea of putting the constructor in it. I think I want to keep this as-is for now, we could always add a trait later.

ryanscherler:
Looking to setup another Jigsaw site and wondering if this might get merged in to use - so excited! 😜

PR #544, opened 2021-02-09
Allow Collection Specific perPage Setting

intrepidws:
Currently, you can set the `perPage` setting either in your template file or as a global setting in config.php. This pull request would allow you to set the `perPage` setting for a specific collection in config.php. The order of preference would be as follows:

1. Within template
2. Collection specific in config.php
3. Global in config.php
4. Default of 10

Note that I don't really love how this looks in the code itself, but it's better than keeping it all on the same line together. One alternative would be a temporary variable, which I don't love either. I'm open to other implementation suggestions.

### DIFF
--- a/src/Handlers/PaginatedPageHandler.php
+++ b/src/Handlers/PaginatedPageHandler.php
@@ -41,16 +41,17 @@ class PaginatedPageHandler
 
     public function handle($file, PageData $pageData)
     {
-        $pageData->page->addVariables($this->getPageVariables($file));
+        $page = $pageData->page;
+        $page->addVariables($this->getPageVariables($file));
+        $collection = $page->pagination->collection;
 
         return $this->paginator->paginate(
             $file,
-            $pageData->get($pageData->page->pagination->collection),
-            $pageData->page->pagination->perPage ?: (
-                $pageData->page->collections->{$pageData->page->pagination->collection}->perPage ?: (
-                    $pageData->page->perPage ?: 10
-                )
-            )
+            $pageData->get($collection),
+            $page->pagination->perPage
+                ?: $page->collections->{$collection}->perPage
+                ?: $page->perPage
+                ?: 10
         )->map(function ($page) use ($file, $pageData) {
             $pageData->setPagePath($page->current);
             $pageData->put('pagination', $page);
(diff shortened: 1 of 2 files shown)


src/Handlers/PaginatedPageHandler.php
```diff
@@ -46,7 +46,11 @@ public function handle($file, PageData $pageData)
         return $this->paginator->paginate(
             $file,
             $pageData->get($pageData->page->pagination->collection),
-            $pageData->page->pagination->perPage ?: ($pageData->page->perPage ?: 10)
+            $pageData->page->pagination->perPage ?: (
+                $pageData->page->collections->{$pageData->page->pagination->collection}->perPage ?: (
+                    $pageData->page->perPage ?: 10
+                )
+            )
```
damiani, line 53: I think it would be a little cleaner to use a couple of temp variables, and then a chained ternary, here:

```php
$page = $pageData->page;
$collection = $page->pagination->collection;
$page->addVariables($this->getPageVariables($file));

return $this->paginator->paginate(
    $file,
    $pageData->get($collection),
    $page->pagination->perPage 
        ?: $page->collections->{$collection}->perPage 
        ?: $page->perPage
        ?: 10
)->map ...
```

damiani:
@intrepidws Great idea! I do think using a couple temp variables make sense here, since that method is so hard to read otherwise. A chained ternary can help, as well.

Could you add a test for this change?

Thanks!

src/Handlers/PaginatedPageHandler.php
intrepidws: Agreed, that's a lot cleaner. I had to swap around the order of the temporary variables to get it to work, but that's in place now.

intrepidws:
Tests forthcoming...

damiani:
Super!


Il giorno sab 27 feb 2021 alle 12:41 Craig Anderson <
notifications@github.com> ha scritto:

intrepidws:
Alright, I have added six new tests, which cover the new functionality as well as expand the test suite to cover some existing functionality as well.

intrepidws:
@damiani Just checking in to see if the tests that I added do the trick?

damiani:
Awesome, thanks @intrepidws!

PR #560, opened 2021-06-13
Give precedence to class based component

avvertix:
Thanks Tighten team for this static site generator.

Following the latest additions for Blade components support I want to propose a change in the anonymous vs class based component precedence giving precedence to class based ones. 

Assuming a Blade template for the component is defined in `_components`, e.g. `_components/image.blade.php`, and the corresponding Component class defined, e.g. `Image extends Component` under the `Components` namespace, inverting the evaluation criteria for class based components with anonymous components allows to add methods and additional attributes to the component that otherwise would be ignored.

This probably covers my specific case in which I need to define an image component whose Blade file is located under `_components/image.blade.php` and whose logic, for processing the image file, is defined in the `Image` component class in the `Components` namespace. This is required for my experiment in adding support for cache busting and responsive images to the blog I'm working on.

### DIFF
--- a/src/View/ComponentTagCompiler.php
+++ b/src/View/ComponentTagCompiler.php
@@ -24,13 +24,13 @@ class ComponentTagCompiler extends BaseComponentTagCompiler
             return $this->aliases[$component];
         }
 
+        if (class_exists($class = $this->guessClassName($component))) {
+            return $class;
+        }
+
         if (Container::getInstance()[Factory::class]
             ->exists($view = "_components.{$component}")) {
             return $view;
-        }
-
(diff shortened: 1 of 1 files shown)


damiani:
This is a great idea, because it seems to bring the component behavior in line with the current version of `Illuminate\View` that Jigsaw is using and overriding, which has drifted a bit since we first implemented this. There are, however, some additional changes in the base `ComponentTagCompiler` class that I'd like to pick up ... would you mind if I add a commit to your PR to include those?

avvertix:
> would you mind if I add a commit to your PR to include those?

Feel free to add what you need @damiani

PR #564, opened 2021-07-19
Allow configuration of Blade view path

bakerkretzmar:
This PR allows users to configure the location of the Blade view path. This facilitates, for example, storing Blade views somewhere other than `source/`. Defaults to `source/` so should be entirely backwards compatible.

### DIFF
--- a/jigsaw-core.php
+++ b/jigsaw-core.php
@@ -67,6 +67,7 @@ $bootstrapFile = $container['cwd'] . '/bootstrap.php';
 
 $container->instance('buildPath', [
     'source' => $container['cwd'] . '/source',
+    'views' => $container['cwd'] . '/source',
     'destination' => $container['cwd'] . '/build_{env}',
 ]);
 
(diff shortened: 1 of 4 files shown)


damiani:
@bakerkretzmar Tests were failing for me because of missing `views` keys in our test setup, where we set `buildPath` in the container directly. 

It also led me to realize that defaulting `views` to `/source` in `jigsaw-core` means that users who set their own `source` directory in `config.php`, but who don't define a `views` key there, would end up with `views` erroneously being set to the default `/source`, rather than the directory that they had defined.

I _think_ this all works now—tests all pass—but I feel like we should add tests for the actual behavior added in this PR, namely that setting a custom `views` directory in config works.

bakerkretzmar:
> @bakerkretzmar Tests were failing for me because of missing `views` keys in our test setup, where we set `buildPath` in the container directly.

I'm seeing this too now... not sure why they worked for me before.

> It also led me to realize that defaulting `views` to `/source` in `jigsaw-core` means that users who set their own `source` directory in `config.php`, but who don't define a `views` key there, would end up with `views` erroneously being set to the default `/source`, rather than the directory that they had defined.

Great point!

> I _think_ this all works now—tests all pass—but I feel like we should add tests for the actual behavior added in this PR, namely that setting a custom `views` directory in config works.

Will do 👍🏻

bakerkretzmar:
@damiani added a test, think that's enough?

damiani:
Love it, great work!

PR #571, opened 2021-07-20
Update Illuminate components to v8

bakerkretzmar:
Does what it says on the box 😄 

@damiani I'm not positive that this won't break anything in any existing sites... thoughts? Might not be a huge deal to leave it until v2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,10 +51,11 @@
     "scripts": {
         "fix" : "php vendor/bin/php-cs-fixer fix --using-cache=no"
     },
-    "prefer-stable": true,
     "config": {
         "sort-packages": true,
         "preferred-install": "dist",
         "optimize-autoloader": true
-    }
+    },
+    "minimum-stability": "dev",
+    "prefer-stable": true
 }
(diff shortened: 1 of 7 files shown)


damiani:
@bakerkretzmar Yeah this worries me a bit, too—I _think_ it should be fine (tests pass, and the collection change preserves pre-v8 behavior), but I'm not sure this upgrade buys us anything of value, so it may not be worth the risk.

bakerkretzmar:
Sounds good, we can shelve it for now. Nice to know at least that it'll be pretty simple 😅

keithbrink:
Just did some work on a PR before I noticed this one 😅

I know this is a little niche, but I would love support for the 8.0 components so I can use the Notion API package in Jigsaw:

https://github.com/5am-code/laravel-notion-api

bakerkretzmar:
We should probably take another look at this, I think we'll have to make this upgrade to be able to support PHP 8.1. See #601.

damiani (approved):

Issue #274, opened 2018-09-27
Jigsaw tries to process dot files and won't build the website


Anthony-Gaudino:
I use VIM to edit my code and VIM creates swap files, for example, `.index.blade.php.swp`.

Jigsaw is trying to process these files and generates errors which cause the website to not build at all.

I think it would be best to not build any dot file or provide an option to do so.

ruman63:
I don't see any point using dot files, when you can simply use underscore files (`_index.blade.php`) which are already supposed to be ignored by the jigsaw.

anthonyterrell:
Hey @Anthony-Gaudino, 

Thank you for reporting this. Are you still running into this bug and Is there a repository we could use to replicate the issue with?

sebsel:
I can confirm that this is very much a thing still.

```
./vendor/bin/jigsaw serve

Building local site
Building files from source...


In Factory.php line 291:

  Unrecognized extension in file: ./source/.index.blade.php.swp
  .


serve [--host [HOST]] [-p|--port PORT] [--no-build] [--] [<environment>]
```

The best way to reproduce is just start a new Jigsaw project and touch a file:

```
mkdir new-jigsaw && cd new-jigsaw
composer require tightenco/jigsaw
./vendor/bin/jigsaw init
touch source/.file.blade.php.swp
./vendor/bin/jigsaw serve
```

Note that it does not happen when I do `touch source/.file.swp`, seems like the `.blade.php` throws it off.

PR #583, opened 2021-09-23
Check for a valid blade extension when handling blade files

jwpage:
Blade files without valid exceptions in will cause an exception to be thrown during building. This is particularly noticeable with editor swap files (eg. vim has `.foo.blade.php.swp`) preventing the site from building. (Also reported in #274)

```
❯ ./vendor/bin/jigsaw build

Building local site
Building files from source...


In Factory.php line 291:

  Unrecognized extension in file: /Volumes/Code/home/jigsaw-test/source/foo.blade.bar.


build [--pretty PRETTY] [-c|--cache [CACHE]] [--] [<env>]
```

This PR makes `BladeHandler::shouldHandle` check that the file has a valid extra blade extension (as defined in `extraBladeExtensions`, or a standard `blade.php` extension. If not, it will be rendered by the default handler.

I'm not sure if this is the best approach, as it would silently let unlisted blade extensions pass through without the user being aware of it, but seems better than breaking the build entirely.

Would it possibly be better if we display a warning to the user notifying them of an invalid blade extension instead?

### DIFF
--- a/src/Handlers/BladeHandler.php
+++ b/src/Handlers/BladeHandler.php
@@ -2,7 +2,6 @@
 
 namespace TightenCo\Jigsaw\Handlers;
 
-use Illuminate\Support\Str;
 use TightenCo\Jigsaw\File\OutputFile;
 use TightenCo\Jigsaw\File\TemporaryFilesystem;
 use TightenCo\Jigsaw\PageData;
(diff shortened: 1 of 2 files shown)


jwpage:
Closing due to inactivity.

PR #584, opened 2021-10-07
Allow merging optional collections file to config file

intrepidws:
I often find myself with a lot of collection definitions inside my `config.php` file. Similar to how the `helpers.php` file allows you to offload helper methods to a separate file, this PR would allow you to offload collection definitions to a separate `collections.php` file.

I have accounted for the following scenarios:

- Collections defined only in `config.php`
- Collections defined only in `collections.php`
- Collections defined in both `config.php` and `collections.php`

I looked into adding tests for this (and also for the currently untested `helpers.php` file) but it would require adding a `collections.php` file to the root directory (likewise with a `helpers.php` file). While I'm not opposed to doing that, I wasn't sure how you might want to handle this.

Note that the root directory doesn't contain a `config.php`, but instead contains a `config.testing.php` (reflecting the environment). The existing `helpers.php` file integration doesn't account for the environment variable like this - and I wanted my `collections.php` integration to work in a similar fashion.

### DIFF
--- a/jigsaw-core.php
+++ b/jigsaw-core.php
@@ -71,7 +71,7 @@ $container->instance('buildPath', [
 ]);
 
 $container->bind('config', function ($c) use ($cachePath) {
-    $config = (new ConfigFile($c['cwd'] . '/config.php', $c['cwd'] . '/helpers.php'))->config;
+    $config = (new ConfigFile($c['cwd'] . '/config.php', $c['cwd'] . '/helpers.php', $c['cwd'] . '/collections.php'))->config;
     $config->put('view.compiled', $cachePath);
     return $config;
 });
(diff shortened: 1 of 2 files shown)


damiani:
Excellent, I'll take a look at what would be required to test this. The `config.testing.php` is there for the snapshot tests, but I'd definitely want to get this (and`helpers.php`) tested.

bakerkretzmar:
Thanks a lot! I like the `collections.php` file idea for keeping things organized if you have lots of collections.

In terms of merging them into your config, I think what you're going for here is already possible by just including that file directly:

```php
<?php

$collections = require_once __DIR__ . '/collections.php';

return [
    'collections' => $collections,
];
```

Not going to merge this at the moment but thanks a lot for the PR!

PR #600, opened 2021-11-27
Allow invokeable event listeners

doekenorg:
Hi, 

There is a slight issue with the `EventBus`. When you add a listener to the event collection, it creates a new collection by merging with a `collect($arguments[0])` call. This has some unforeseen side effects when you use an invokable event listener with an internal object.

```php
$events->afterCollections(new class {
    private $object;

    public function __construct()
    {
        $this->object = (object) [];
    }

    public function __invoke(Jigsaw $jigsaw) {
        dump($jigsaw);
    }
});
```

This example will currently throw an error:

```bash
PHP Fatal error:  Uncaught Error: Call to undefined method stdClass::handle() in /Users/doekenorg/code/doekenorg/vendor/tightenco/jigsaw/src/Events/EventBus.php:33
```

This update will fix that problem, by just appending the first argument to the collection that already exists. No merging required.

### DIFF
--- a/src/Events/EventBus.php
+++ b/src/Events/EventBus.php
@@ -2,6 +2,7 @@
 
 namespace TightenCo\Jigsaw\Events;
 
+use Illuminate\Support\Arr;
 use TightenCo\Jigsaw\Jigsaw;
 
 class EventBus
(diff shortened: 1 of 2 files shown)


doekenorg:
Sorry about that; I somehow completely missed the unit tests. Luckily they exists. 

So the actual problem is that `collect()` just casts the input to an array; which is why this object is interpreted incorrectly. 
To fix this, I wrap the listeners in an array, if they are not already. 

Also added a unit test for this use case.

bakerkretzmar:
@doekenorg thanks for updating this, I was just about to comment the same thing 😅

bakerkretzmar (approved):

doekenorg:
@bakerkretzmar nice. I knew there probably was a nicer way; but I'm not thát familiar with the framework.

bakerkretzmar:
Thank you! 🙌🏻

PR #630, opened 2022-05-08
Adding the check if the function exists.

jeffersonsimaogoncalves:


bakerkretzmar:
Thanks, I like this but I just want to understand its implications. @damiani do you have any idea how we were getting away with redefining these before??

damiani:
I don't see a downside to this, but it seems unnecessary — there's no conflict currently because `public_path` is defined in `Illuminate/Foundation` (the Laravel app core), which Jigsaw does not include as a dependency. @jeffersonsimaogoncalves I'm curious, did you install Laravel _inside_ your Jigsaw project?

PR #635, opened 2022-05-31
feat: add event bus type hints

owenvoke:
I was recently adding some event listeners to a Jigsaw site and noticed that PHPStan complains that these methods don't exist on the `EventBus` instance. 👍🏻

I also added the `Collection` type to the 3 properties as they are always that type (but I'm happy to remove that).


bakerkretzmar (approved):
Thanks!

PR #643, opened 2022-07-14
Upgrade tailwindcss to v3

### LINKED ISSUE #607: Upgrade to Tailwind 3 creates infinite loop in 'npm run watch'
Bit of a strange one - when I try to upgrade tailwind to 3.0, then run `npm run watch` the terminal window shows infinite rebuilding, and there is a 'cache' directory that appears and disappears in the root.

I tried it on the vanilla jigsaw build and the blog template, and it happened on both.

(On a MBP and not using anything like docker - if that helps)

I have a screen recording of this if it's any use.

khatabwedaa:
resolve #607

### DIFF
--- a/stubs/site/package.json
+++ b/stubs/site/package.json
@@ -9,8 +9,8 @@
     "devDependencies": {
         "laravel-mix": "^6.0.39",
         "laravel-mix-jigsaw": "^1.5.0",
-        "postcss": "^8.2.4",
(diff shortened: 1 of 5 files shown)


damiani:
@khatabwedaa This is fantastic, thanks! I added `blade.md` to the Tailwind content array (thanks to @jasonlbeggs for pointing that out). Appreciate it!

PR #651, opened 2022-08-30
Improve serve command option handling

paulofreitas:
This allows us to call `jigsaw serve` command with `-q` or `-vv` options and it will work just like `jigsaw build` command. 👍

Currently if you use a custom build path and run `jigsaw serve -q` it does not ask for a confirmation nor continue - it's kinda broken. 😄


damiani:
Thanks @paulofreitas !

PR #656, opened 2022-09-06
Add Browsersync

piercemcgeough:
Adding browsersync for base app


bakerkretzmar:
Thanks for the PR! We intentionally keep the basic site stub as simple as possible and Browsersync has been flaky in the past, so we're going to pass on adding this right now. We did notice that the docs reference Browsersync always being available, we'll get that fixed.

PR #661, opened 2022-09-30
Update `illuminate/*` components to v9

bakerkretzmar:
This PR upgrades Jigsaw's dependencies on several of Laravel's Illuminate components to version 9, released earlier this year. This upgrade necessitates dropping support for PHP 7.3 and 7.4, which is inconvenient but had to happen eventually. Laravel 9 starts using the `mixed` type, which is only available in PHP 8. I played around with replacing it with a union of several other types, but... union types are also only available in PHP 8.

I don't think this is a breaking change, because Jigsaw users on PHP 7.3 and 7.4 will not be able to update at all. There are underlying changes that _could_ be breaking, like the update to Flysystem 3, but this feels unlikely to me... @damiani thoughts?


damiani:
Agreed, I think this should be safe 👍

damiani (approved):

PR #662, opened 2022-09-30
Refactor container bootstrapping to use service providers

bakerkretzmar:
This PR refactors Jigsaw's internal setup and service container bootstrapping to look a lot more like Laravel's Artisan CLI setup, including using our own custom Container class and introducing ServiceProvider classes. This starts laying the groundwork for some _very_ interesting future possibilities, including:

- Cleaner userland events/hooks/extensions with user ServiceProviders.
- Powerful Jigsaw packages/plugins with auto-discovery.
- Improved testability of Jigsaw itself and Jigsaw sites and plugins.

Major plot points of this PR:

- Create Jigsaw's own `Container` class, extending `Illuminate\Container\Container`, and do environment/config bootstrapping and service provider setup there.
- Remove the `jigsaw-core.php` file and perform most container setup in several new ServiceProvider classes, like `ViewServiceProvider` and `CollectionServiceProvider`.
- Use the real Illuminate event dispatcher.
- Add a new global `app()` helper to retrieve the container instance.
- Add 'booting' and 'booted' callbacks in the container (not particularly useful until we add user ServiceProviders but still nice).
- Add new container path helpers: `app()->path()` and `app()->cachePath()` (more to come).

Other changes worth mentioning—technically public-facing, but so minor or obscure that I don't think we need to worry about them breaking anything:

- The global application container is now an instance of `TightenCo\Jigsaw\Container` instead of `Illuminate\Container\Container`. Our container extends the Illuminate one though, so e.g. an `instanceof Container` check would still pass.
- We are registering a lot of container bindings in a different order than we used to, but I'm pretty sure it would be extremely difficult or even impossible for this to matter, because there wasn't really any way to depend on this order previously.
- The `TightenCo\Jigsaw\Events\FakeDispatcher` class was removed.

See tightenco/jigsaw-v2#1.


bakerkretzmar:
Just discovered another potential hiccup here—because `jigsaw-core.php` wasn't namespaced, all its variables were declared globally, making them available in `$GLOBALS`. After this PR they'll be missing from `$GLOBALS`.

damiani (approved):

PR #665, opened 2022-10-21
Allow filename.md.blade.php

haringsrob:
When working with blade, editing is more easy when using .blade.php.

This addition will allow us to continue using markdown, but also having the benefit of autocomplete in editors.

### DIFF
--- a/src/View/ViewRenderer.php
+++ b/src/View/ViewRenderer.php
@@ -19,6 +19,7 @@ class ViewRenderer
         'blade.md' => 'blade-markdown',
         'blade.mdown' => 'blade-markdown',
         'blade.markdown' => 'blade-markdown',
+        'md.blade.php' => 'blade-markdown',
     ];
     private $bladeExtensions = [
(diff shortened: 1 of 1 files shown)


bakerkretzmar:
What exactly does this let you do? Use Markdown but have your editor highlight the file as if it's Blade? My gut feeling is that that's something people should configure individually in their editor (e.g. treating `.blade.md` files as PHP).

haringsrob:
> What exactly does this let you do? Use Markdown but have your editor highlight the file as if it's Blade? My gut feeling is that that's something people should configure individually in their editor (e.g. treating `.blade.md` files as PHP).

That is indeed what I have done for now and works. This pr would allow that without any configuration requirement.

bakerkretzmar:
@haringsrob thanks a lot for the PR! Going to pass on this right now. I think this is a specific enough preference and has an easy enough workaround (setting up your editor as discussed above) that it's not necessary to include in Jigsaw itself. It would also be a breaking change for anyone who already has `.md.blade.php` files.

PR #668, opened 2022-11-11
Improve error handling

bakerkretzmar:
This PR makes some long-overdue improvements to Jigsaw's error handling. The code is mostly copied from Laravel and tweak slightly, and lots of Laravel functionality was removed so that we're only implementing what we absolutely need for now.

#### Highlights

- Adds an Artisan-like exception handler to give us more control over error handling.
- Installs [Collision](https://github.com/nunomaduro/collision) and renders errors with it, so they're now much prettier, show stack traces, and include basic syntax highlighting.
- Installs [Laravel Ignition](https://github.com/spatie/laravel-ignition) and maps Blade errors through its custom handler/compiler, so their stack traces display the original **uncompiled** Blade templates.

#### Demo

Here are before and after comparisons for a few possible error scenarios:

### 1. Undefined variable in Blade template

```blade
{{ $test }}
```

**Before**

Build succeeds (this is bad, it should fail), no error output in the console and no indication that anything is wrong. Additionally, the missing variable is rendered as an error visible in the HTML of the built site:

```html
Warning: Undefined variable $test in /Users/jacob/Code/tighten/tightendotco/cache/914bb3c0cbe5c1750461e191b6c1f4a331a98ef3.php on line 2
```

**After**

Build fails with a nice-looking and detailed error message, including a stack trace pointing to the source of the error in the original, **uncompiled** Blade template:

Screen Shot 2022-11-18 at 12 38 29 PM

### 2. Invalid expression / syntax error in Blade template

```blade
{{ aaaaa ->jnpa af.asdf }}
```

**Before**

Build fails with a short, ugly error message (in this case it indicates correctly where the error came from, but I've seen it get that wrong or not provide that info at all many times):

Screen Shot 2022-11-18 at 12 06 40 PM

**After**

Build fails with a pretty, detailed error message and stack trace:

Screen Shot 2022-11-18 at 12 41 54 PM

### 3. PHP deprecation

```blade
{{ str_contains(null, 'f') }}
```

**Before**

Build succeeds, no error messages, and the deprecation warning is rendered into the output of the site:

```html
Deprecated: str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated in /Users/jacob/Code/tighten/tightendotco/cache/5ca2f78ccab5990f8ff2686eb4263c891fd085e1.php on line 11
```

**After**

Build succeeds, but displays warnings in the console, and the error is **not** rendered into the built HTML:

Screen Shot 2022-11-18 at 3 17 28 PM

### 4. Exception thrown manually

```php
// inside a collection helper in config.php
throw new RuntimeException('Uh oh...');
```

**Before**

Screen Shot 2022-11-18 at 11 54 36 AM

**After**

> The original source line of the exception is not visible in this screenshot. This is a small trade-off of this PR—because we're now rendering _all_ exceptions with a lot more detail, some specific cases like this one will no longer display exactly what we need initially. The source line we're looking for here (in `config.php`) is frame <span>#</span>3, which becomes visible if you re-run the build verbosely (`-v`). Eventually we could fix this with our own custom exception renderer that shows more stack frames, but that feels like overkill for now.

Screen Shot 2022-11-18 at 12 05 21 PM

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,9 @@
         "illuminate/view": "^9.8",
         "michelf/php-markdown": "^1.9",
         "mnapoli/front-yaml": "^1.5",
+        "nunomaduro/collision": "^6.0",
         "symfony/console": "^5.4 || ^6.0",
+        "symfony/error-handler": "^5.0 || ^6.0",
         "symfony/finder": "^5.3.7 || ^6.0",
         "symfony/process": "^5.0 || ^6.0",
         "symfony/var-dumper": "^5.0 || ^6.0",
@@ -33,7 +35,8 @@
         "friendsofphp/php-cs-fixer": "^3.11",
         "mikey179/vfsstream": "^1.6.10",
         "mockery/mockery": "^1.4",
-        "phpunit/phpunit": "^9.3.3"
+        "phpunit/phpunit": "^9.3.3",
+        "ramsey/composer-repl": "^1.4"
     },
     "autoload": {
         "psr-4": {
@@ -55,8 +58,11 @@
         "format" : "php-cs-fixer fix --verbose"
     },
     "config": {
+        "sort-packages": true,
         "optimize-autoloader": true,
-        "sort-packages": true
+        "allow-plugins": {
+            "ramsey/composer-repl": true
+        }
     },
     "minimum-stability": "dev",
     "prefer-stable": true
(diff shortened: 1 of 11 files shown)


damiani (approved):

damiani:
Love this! 🎉

mattstauffer:
WHOA. Love this!

husseinalhammad:
Awesome stuff 🚀

PR #677, opened 2023-02-03
Add support for Laravel 10

jasonmccreary:
This supersedes [#676](https://github.com/tighten/jigsaw/pull/676) by ensuring both Laravel 9 and Laravel 10 are supported. 

In addition, it sets the minimum requirement for Symfony dependencies to `^6.0` as that was the minimum required by Laravel 9.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "illuminate/support": "^9.0 || ^10.0",
         "illuminate/view": "^9.8 || ^10.0",
         "michelf/php-markdown": "^1.9 || ^2.0",
-        "mnapoli/front-yaml": "^1.5 || ^2.0",
+        "mnapoli/front-yaml": "^2.0",
         "nunomaduro/collision": "^6.1",
         "spatie/laravel-ignition": "^1.6 || ^2.0",
         "symfony/console": "^6.0",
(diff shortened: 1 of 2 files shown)


jasonmccreary:
The failing build for PHP 8.3 is because of the bump in `mnapoli/front-yaml` to be compatible with Symfony 6. One of its underlying dependencies has a `<8.3` PHP requirement.

Since 8.3 is not officially released yet, I am not going to worry about this as it should resolve itself in time. However, you may want to remove PHP `8.3` from the matrix build until then.

jasonmccreary:
The failing "Format" job is due to a restriction of the Git remote. It's currently `origin`. However, since I am an outside contributor, my branch is on a different remote. This job should be updated to properly pull from other remotes, or skipped when the branch does not exist on `origin`.

bakerkretzmar:
I'm not sure we can upgrade `mnapoli/front-yaml` without tagging a new major version. Is that directly related to supporting Laravel 10 or could we just add that and only touch other dependencies we absolutely have to?

owenvoke:
I think Front YAML has to be updated as v1 doesn't support Symfony 6 ☹️

damiani:
Superseded by #702

PR #686, opened 2023-05-12
Allow setting collections using a function

leopoletto:
This PR addresses the issue https://github.com/tighten/jigsaw/issues/572

It allows us to set the collection using a function.

It means that now it is possible to get the collections and entries from an external API dynamically without having to expressly set the collections.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -109,14 +109,12 @@ class TestCase extends BaseTestCase
         $this->app->consoleOutput->setup($verbosity = -1);
         $this->app->config = collect($this->app->config)->merge($config);
 
-        $collections = value($this->app->config->get('collections'));
-
-        if ($collections) {
+        if ($collections = value($this->app->config->get('collections'))) {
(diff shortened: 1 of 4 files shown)


tests/TestCase.php
```diff
@@ -108,6 +108,13 @@ public function buildSite($vfs, $config = [], $pretty = false, $viewPath = '/sou
     {
         $this->app->consoleOutput->setup($verbosity = -1);
         $this->app->config = collect($this->app->config)->merge($config);
+
+        if ($collections = value($this->app->config->get('collections'))) {
+            $this->app->config->put('collections', collect($collections)->flatMap(function ($value, $key) {
+                return is_array($value) ? [$key => $value] : [$value => []];
+            }));
+        }
```
bakerkretzmar, line 116: Why is this necessary?

leopoletto (commented):
Revert all these whitespace/formatting changes please.

leopoletto:
@bakerkretzmar Sorry for not answering and fixing it. Thanks for merging. 💯

PR #688, opened 2023-05-26
Use real filesystem in tests

bakerkretzmar:
This PR updates all tests to use the real filesystem instead of a virtual one, and uninstalls `mikey179/vfsstream`. This brings the test environment/behaviour closer to how Jigsaw actually runs in real environments, and it makes debugging and inspecting failing tests significantly easier.

Kind of similarly to Laravel's `Storage::fake()` (which only cleans up previous tests' files at the beginning of the next test run), this PR only cleans up test files if all tests passed, so if anything goes wrong the test files are preserved and we can review them.

I also added some cleaner setup helper methods, but to keep this PR from getting too noisy I didn't replace the old ones everywhere yet.

The performance hit here (I _think_ using the real filesystem is slower) is negligible.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -103,6 +103,9 @@ class TestCase extends PHPUnit
         return "{$this->tmp}/{$path}";
     }
 
+    /**
+     * @deprecated Use createSource instead.
+     */
     protected function setupSource($source = [])
     {
         $this->createSource(['source' => $source]);
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -61,9 +61,8 @@ class TestCase extends PHPUnit
     {
         mkdir($this->tmp = __DIR__ . '/fixtures/tmp/' . static::haiku());
 
-        mkdir("{$this->tmp}/cache");
-        mkdir("{$this->tmp}/cache/views");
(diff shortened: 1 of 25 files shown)


tests/TestCase.php
```diff
@@ -43,20 +55,22 @@ protected function setUp(): void
...
-        $this->filesystem = new Filesystem();
-        $this->tempPath = $this->app->cachePath();
-        $this->prepareTempDirectory();
+    }
+
+    protected function createTmp(): void
+    {
+        mkdir($this->tmp = __DIR__ . '/fixtures/tmp/' . static::haiku());
+
+        // TODO this creates Jigsaw's cache directory in the root of this repo
```
damiani, line 64: I think this comment can be removed, right?

damiani (commented):
A thing of beauty!

tests/TestCase.php
bakerkretzmar: I don't think so—that was already the behaviour before this PR and the Blade compiler relies on that directory existing. Eventually I want to fix it but that might require more fiddling with how Jigsaw determines its 'base path' and working directory.

But actually only 5 tests fail without it so let me see!

PR #701, opened 2024-03-18
fixing PHPUnit BC break on base TestCase

njames:
I am updatating jigsaw's packages to get to a point where I can add laravel/prompts
PHPUnit 11 seems to have a BC break - anyone know about this? 

TBC This PR is not ready yet I am just trying to get other eyes on the PHPUnit issue.

```
An error occurred inside PHPUnit.

Message:  Too few arguments to function PHPUnit\Framework\TestCase::__construct(), 0 passed in /Users/nigeljames/OS/jigsaw/tests/TestCase.php on line 31 and exactly 1 expected
Location: /Users/nigeljames/OS/jigsaw/vendor/phpunit/phpunit/src/Framework/TestCase.php:224
```

I have resolved this with https://stackoverflow.com/a/66676166 and I am now working on other updates with Symphony packages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 vendor/
+.idea/
 .php-cs-fixer.cache
 .phpunit.result.cache
 composer.lock
--- a/composer.json
+++ b/composer.json
@@ -18,30 +18,31 @@
         }
     ],
     "require": {
-        "php": "^8.0.2 <8.4",
-        "illuminate/collections": "^9.0",
-        "illuminate/console": "^9.21.1",
-        "illuminate/container": "^9.0",
-        "illuminate/filesystem": "^9.0",
-        "illuminate/support": "^9.0",
-        "illuminate/view": "^9.8",
+        "php": "^8.3",
+        "illuminate/collections": "^11.0.7",
+        "illuminate/console": "^11.0.7",
+        "illuminate/container": "^11.0.7",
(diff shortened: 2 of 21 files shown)


njames:
OK so now `composer update` and `jigsaw init`  will run but the tests are failing still. 

I have fixed the first couple of PHPUnit and Symfony errors I have come across and now have a markdown error. 

I am keeping my commits granular and hopefully this helps. 

CC @bakerkretzmar

bakerkretzmar:
@njames thanks a lot for digging into this 🙏🏻 

If I remember correctly, strictly speaking we can't upgrade the mnapoli markdown packages without a major version bump to Jigsaw itself, because there are backwards-incompatible changes to some public interfaces. I can take a closer look later but that looks like it's the issue with at least some of the failing tests here.

If we decide we don't mind bumping Jigsaw to v2 just to do those upgrades, then this PR looks like a great start and I'd be happy to jump in and help get things passing this week.

damiani:
I've run through and made a handful of PHPUnit fixes, and I currently have all tests passing; the only markdown-related error was a missing return type. I'll take a closer look to the `mnapoli` changes that @bakerkretzmar mentioned, to see if there are indeed any breaking changes that would require a major version bump for Jigsaw.

damiani:
Addressed in #702

PR #706, opened 2024-05-15
Fix Format Action for Forked Repos

scottcharlesworth:
Addresses an issue with the `format.yml` GitHub Action workflow where it fails for pull requests from forked repositories. It adds the `repository` parameter, ensuring the correct branch from the fork is checked out.

Without this PRs fail checks with: A branch or tag with the name 'xyz-etc' could not be found


damiani:
Thanks again!

PR #708, opened 2024-06-25
Fix dynamic components

### LINKED ISSUE #707: <x-dynamic-component /> support?
Is the Laravel documented support for [Dynamic Components](https://laravel.com/docs/11.x/blade#dynamic-components) is expected to work in Jigsaw?

An exception is raised when trying it out and it's unclear if it is local to the project or broadly unsupported in Jigsaw.

```php
$componentType = 'something-kebab-case';

<x-dynamic-component :component="$componentType" />
```

```bash
   Spatie\LaravelIgnition\Exceptions\ViewException 
  Target [Illuminate\Contracts\Foundation\Application] is not instantiable.
```

In this example, a component is confirmed as working when expressed explicitly, so it is not a missing component issue.

This works just fine.

```php
<x-something-kebab-case />
```

The error _feels_ architectural. 

I also noticed there is some reference to dynamic component support in the Jigsaw providers, but there are also some code comments and TODOs. 

https://github.com/tighten/jigsaw/blob/114ed1650af48ea9c4ce333d08cc45c97ad5b928/src/Providers/ViewServiceProvider.php#L61-L72

Is this expected to be supported and unfinished?

bakerkretzmar:
Fixes `<x-dynamic-component>` Blade components, which don't work but should have been supported (and might have worked at some point in the past).

All this does is override the `DynamicComponent` component class so that it uses Jigsaw's customized `ComponentTagCompiler` instead of Laravel's. I didn't even change anything in `DynamicComponent`, just the fact that we redeclare it picks up the new compiler because of the namespace it's in.

There might be more nuance to explore here, this is super basic and just makes the happy path example in the included snapshot test work as expected.

Closes #707.

### DIFF
--- a/tests/snapshots/dynamic-component_snapshot/index.html
+++ b/tests/snapshots/dynamic-component_snapshot/index.html
@@ -1,0 +1,7 @@
+<div class="alert alert-danger">
+    <h3>This is the component</h3>
+    <h4>Named title slot: Title</h4>
+    <hr>
+        Slot
+    <hr>
+</div>
--- a/src/Providers/ViewServiceProvider.php
+++ b/src/Providers/ViewServiceProvider.php
@@ -2,7 +2,6 @@
 
 namespace TightenCo\Jigsaw\Providers;
 
-use Illuminate\View\DynamicComponent;
 use Illuminate\View\Engines\CompilerEngine;
 use Illuminate\View\Engines\EngineResolver;
 use Illuminate\View\Engines\FileEngine;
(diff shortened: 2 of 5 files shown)


bakerkretzmar:
Leaving the formatting action broken here because it changes a ton of files. I'll fix it separately.

stevenmaguire:
@bakerkretzmar thanks for getting this going. I think this is the most prudent strategy. 

> There might be more nuance to explore here, this is super basic and just makes the happy path example in the included snapshot test work as expected.

The basics working is a good sign. If you're looking for most use cases, my specific use case is that I am using "class backed" components in the `\Components` jigsaw project namespace. I know that there is a logical fork in the component detection logic - one to source the path of the php class that backs the component and one to source the path of the blade template if the first fails. Perhaps ensuring that use case is handled in your solution would help build confidence?

damiani (approved):

damiani:
Fantastic, thanks @bakerkretzmar !!

stevenmaguire:
Thank you both!

tests/snapshots/dynamic-component_snapshot/index.html
```diff
@@ -0,0 +1,7 @@
+<div class="alert alert-danger">
+    <h3>This is the component</h3>
+    <h4>Named title slot: Title</h4>
+    <hr>
+        Slot
```
iSWORD, line 5: just curious, do we have access to `$page` here? if so, this is revolutionary!

PR #726, opened 2025-03-10
Update stubs

nicodevs:
Hey @gcavanunez,

I prepared this PR to be merged into yours. It essentially updates the stubs Jigsaw comes with.  

In your PR, you laid out a set of steps to use Vite with Jigsaw. These steps include:  

- Add `"type": "module"` to your `package.json`  
- Create file `vite.config.js`  
- Add scripts to your `package.json`  
- Install Tailwind  
- Tweak `postcss.config.js`  

With this PR merged into yours, those steps will already be done because the `init` command will copy these new stubs.  

Now, I am experiencing issues with hot reloading. On my local setup, it is not always hot reloading as expected. But following your steps, hot reloading works, so I wonder if the issue is in a specific _version_ of the dependencies updated in `package-lock.json`.  

So I'd recommend creating an app and using this branch:  

```sh
composer require tightenco/jigsaw:dev-nd/add-vite-2
```  

...and see if your local hot module reloading works as expected, to check if you experience the same issue or not. If you merge this into yours, feel free to squash the commits.

Thank you!


stubs/site/source/hot
gcavanunez, line 1: I think this guy is maybe causing the issue, we should add it to the gitignore

gcavanunez (approved):
This is great! after removing the `hot` file everything worked as expected hot reloading and all

PR #731, opened 2025-04-15
Add support for environment-related Blade directives

osbre:


osbre:
Actually, this should work differently. The build command already detects the environment based on the presence of `config.production.php`, not the config value itself. Marking as a draft while updating.

damiani:
This is great, thanks!

PR #754, opened 2025-11-27
Keep the leading slash in pretty output path for index files

lancechentw:
Found that `$this->outputPathResolver->link()`  would generate path without a leading slash for index files, e.g. index.blade.php, index.html, index.md, when using pretty urls (which is the default!)

This causes problem when we have `source/blog/index.blade.php` instead of `source/blog.blade.php`. `link()` would return `blog` for the former file, and return `/blog` for the latter file.

For example, GenerateSitemap.php from the blog template would generate the following if having `source/blog/index.blade.php`
```diff
   <url>
-   <loc>https://jigsaw-blog-template.tighten.co/blog</loc>
+   <loc>https://jigsaw-blog-template.tighten.coblog</loc>
    <lastmod>2025-11-27T07:21:16+00:00</lastmod>
    <changefreq>daily</changefreq>
   </url>
```

I have traced back to the commit https://github.com/tighten/jigsaw/commit/7aa603ab2d5177eaf13a2c4e46b6512789d925f4#diff-76e91e6d5ebc8f831ce78c19c83441aa3b719af6a3d4098823f9195a639ec285R11 that introduced the change, and I think it was merely a bug, `ltrim()` should be applied right before `$this->trimPath()`.

### DIFF
--- a/src/PathResolvers/PrettyOutputPathResolver.php
+++ b/src/PathResolvers/PrettyOutputPathResolver.php
@@ -11,7 +11,7 @@ class PrettyOutputPathResolver
                 return '/' . leftTrimPath(trimPath($path) . '/') . trimPath($prefix . '/' . $page) . '/';
             }
 
-            return '/' . leftTrimPath(trimPath($path) . '/') . '/';
+            return '/' . leftTrimPath(trimPath($path) . '/');
         }
 
         if ($type === 'html' && $name !== 'index') {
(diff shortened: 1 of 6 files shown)


nicodevs:
Confirmed this works: built a local site with `source/blog/index.blade.php` and `source/blog/first-post/index.blade.php`, paths come out `/blog/` and `/blog/first-post/` as expected. Before this fix they'd be missing the leading slash, which breaks sitemap URLs etc.

Also added some unit tests to `FilePathTest.php` for root, single-level, nested, and paginated index files.

PR #765, opened 2026-02-13
Add output markdown option

nicodevs:
In this AI era, many websites offer a *Markdown* version of their pages using the same URL with a `.md` extension (like this: [https://laravel.com/docs/12.x/ai-sdk.md](https://laravel.com/docs/12.x/ai-sdk.md)) to make them easier for LLMs to consume.

This PR adds a collection option to Jigsaw to automatically generate these `.md` files.

## Usage

In `config.php`:

```diff
  'collections' => [
      'posts' => [
          'path' => 'blog/{filename}',
+          'output_markdown' => true,
      ],
  ],
```

This produces both `blog/my-post/index.html` (or `blog/my-post.html`) and `blog/my-post.md` for each collection item.

`output_markdown=true` is in fact a shortcut of the longer config options, which are detailed below:

```php
'output_markdown' => ['title_as_h1' => true, 'table_of_contents' => true]
```

## Title as H1

By default, the title of the resource is used as the level-one heading in the generated Markdown. For this resource:

```md
---
title: How to Install Jigsaw
---

Hello! Today...
```

…the output will be:

```md
# How to Install Jigsaw

Hello! Today...
```

If the user does not want this behavior, they can opt out by setting:

```php
'output_markdown' => ['title_as_h1' => false],
```

## Table of Contents

Additionally, by default, the body headings are parsed to build a table of contents at the top of the generated Markdown. For example, given the following document:

```md
---
title: Test
---

## Introduction

This is the intro.

## Installation

You can install the Laravel AI SDK via Composer.
```

The output will be:

```md
# Test

- [Introduction](#introduction)
- [Installation](#installation)

<a name="introduction"></a>
## Introduction

This is the intro.

<a name="installation"></a>
## Installation

You can install the Laravel AI SDK via Composer.
```

If the user does not want this behavior, they can opt out by setting:

```php
'output_markdown' => ['table_of_contents' => false],
```

## Implementation Details

When `output_markdown` is enabled, `CollectionItemHandler` now returns two `OutputFile` instances to the build output: the original file and a raw Markdown version.

To add the title and the table of contents to the markdown we have a new class, `﻿PlainMarkdownBuilder`.

No new dependencies are introduced.

> [!NOTE]
> Note for reviewers: many of the listed “Files changed” reflect linting-only modifications.

### DIFF
--- a/src/Handlers/CollectionItemHandler.php
+++ b/src/Handlers/CollectionItemHandler.php
@@ -5,8 +5,8 @@ namespace TightenCo\Jigsaw\Handlers;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
-use TightenCo\Jigsaw\File\OutputFile;
 use TightenCo\Jigsaw\Builders\PlainMarkdownBuilder;
+use TightenCo\Jigsaw\File\OutputFile;
 use TightenCo\Jigsaw\Parsers\FrontMatterParser;
 
 class CollectionItemHandler
--- a/src/Providers/CollectionServiceProvider.php
+++ b/src/Providers/CollectionServiceProvider.php
@@ -2,6 +2,7 @@
 
 namespace TightenCo\Jigsaw\Providers;
 
+use TightenCo\Jigsaw\Builders\PlainMarkdownBuilder;
 use TightenCo\Jigsaw\Collection\CollectionPaginator;
 use TightenCo\Jigsaw\CollectionItemHandlers\BladeCollectionItemHandler;
 use TightenCo\Jigsaw\CollectionItemHandlers\MarkdownCollectionItemHandler;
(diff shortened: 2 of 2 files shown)


tests/CollectionItemTest.php
```diff
@@ -245,6 +245,103 @@ public function collection_item_page_metadata_contains_modified_time()
...
+        $config = collect(['collections' => [
+            'collection' => [
+                'path' => 'collection/{filename}',
+                'output_markdown' => true,
+            ],
+        ]]);
+
+        $yaml_header = implode("\n", ['---', 'extends: _layouts.collection_item', 'section: content', '---']);
+
+        $files = $this->setupSource([
```
tonysm: I think `setupSource` is deprecated, right? We should use `createSource` instead

nicodevs: Yeah, I followed the rest of the codebase. Perhaps we need to change all occurrences of `setupSource` in another PR, agree?

PR #769, opened 2026-02-24
Optimize manifest loading.

midnight-wonderer:
I use Vite for image hash generation, too.
It would be great if `manifest.json` wasn't read repeatedly during the build.
So, I've implemented static caching for the manifest content.

### DIFF
--- a/src/Support/Vite.php
+++ b/src/Support/Vite.php
@@ -30,12 +30,7 @@ class Vite
         }
 
         $manifestPath = source_path($assetPath . '/manifest.json');
-
-        if (! file_exists($manifestPath)) {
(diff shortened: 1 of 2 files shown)


nicodevs:
This PR caches the parsed `manifest.json` per `Vite` instance so it's only read from disk once per build, rather than on every `url()` call. Useful when `url()` is invoked for many assets (e.g. image hashing).

I've made a change so the cache is stored as an instance property (keyed by manifest path) rather than a `static` variable, so each `Vite` instance gets a fresh read, with no stale data across builds or test cases.

Also added test coverage for the exception paths, custom `$assetPath`, repeated calls, and the stale-cache issue.

